PR #4, opened 2015-03-13
Optimized the block and offset calculations

IDisposable:
Also removed all redundant dotting of memoryManager and fixed end calculation overflow issue.
Mostly this affects the ability of the JIT to know that the backing fields are not changing while we're inside method bodies so better code can result. The overflow issue is that  the new end position calculation was done against ints, not longs, so it could silently overflow.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -140,7 +140,7 @@ namespace Microsoft.IO
         /// </summary>
         /// <param name="memoryManager">The memory manager</param>
         public RecyclableMemoryStream(RecyclableMemoryStreamManager memoryManager)
-            : this(memoryManager, null)
+            : this(memoryManager, null, 0, null)
         {
 
         }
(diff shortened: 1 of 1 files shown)


doubleyewdee:
I see what you're saying now (though I would still prefer benchmarks). My preferred solution to this is to make the manager readonly. I don't think the way we do Dispose is necessary. I'll discuss it with Ben as well and possibly make that change myself.

I would also note I want to avoid optimizing for (e.g.) just one JIT compiler too. I think in this case the mentality is fundamentally sound and should work for any reasonable JIT compiler, just want to call that out.

IDisposable:
Understood completely. The (hard to judge) benchmarks I had for the degenerate cases of lots of small writes were showing a less than 1% improvement, I'm mostly just an advocate for doing the right thing when dealing with low-level code.  To be clear, the improvement from the style changes is **not** JIT specific, it helps all of them (and actually helps the older JITters more).

I really want to make sure that if you decide to not accept this patch that you don't lose the overflow bug fix and the redundant ```Check_Disposed``` call

doubleyewdee:
So actually I went back to look at this today and 'memoryManager' is already readonly. I made a few other fixes which I'll push in, and I'll bring along the block offset calculation changes as well.

doubleyewdee:
Pending add via https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream/pull/9

IDisposable:
Looks good, thanks for doing the patch up.  Just one note, the call to ```this.Position``` instead of ```this.position``` will result in a double-call of ```this.Check_Position()```

PR #8, opened 2015-03-19
Introduce IRecyclableMemoryStreamManager, fixes #6

### LINKED ISSUE #6: Create abstract interface for RecyclableMemoryStreamManager
Please create abstract interface for RecyclableMemoryStreamManager public surface to facilitate unit-testing.

shatl:


benmwatson:
Can you explain more why you think this is needed? Is it for easier unit testing? If so, what in the existing library makes this hard to use in a unit test library? There are no other external dependencies, and it should be easily used. If not, there are mocking libraries for that sort of thing.

shatl:
Yes, this is for unit-testing too.
Currently RecyclableMemoryStreamManager cannot be mocked just because there are no virtual methods. Basically, this can be mitigated by extracting code interacting with stream to separate method accepting MemoryStream as parameter and exposing it to unit-test, but it is not necessary the best  option. 

Second reason would be dynamic proxies (Castle project) which again requires interface or virtual methods.

doubleyewdee:
I believe Ben was referring to using Moq (or similar frameworks which use reflection to generate proxies). Adding interface behavior is not actually free (http://blogs.msdn.com/b/vancem/archive/2006/03/13/550529.aspx) so we want to consider the cost/benefit carefully here.

I also remain confused about why you need this for unit testing, beyond using something like Moq, what would you do with a mocked manager that would facilitate better testing in this situation?

shatl:
Thanks for pointing to this article, but basically interface dispatching overhead is pretty small. Considering real-life scenario, time spent in call to .GetStream() (either direct or via interface) would be not significant with regards to time spent in writing to requested stream.

I can think of several test scenarios, but most important would be tests for handling of exceptions occurred while writing to stream. Currently I implemented local wrapper for mocking and dependency isolation purposes.

benmwatson:
Yes, interface dispatch isn't really the issue here. I'm still not convinced of the need for the interface. It understand it could be useful, but it seems too narrow in scope to merit introduction at this time.

Issue #5, opened 2015-03-13
Exception is thrown on when stream is disposed multiple times


shatl:
Do not throw exception if stream is disposed more than once because this violates default Dispose() behavior.

According to MSDN https://msdn.microsoft.com/en-us/library/system.idisposable.dispose.aspx 

>If an object's Dispose method is called more than once, the object must ignore all calls after the first one. The object must not throw an exception if its Dispose method is called multiple times.

jbattermann:
+1. This is easy to reproduce and also a bit unexpected e.g. when using a StreamReader with a RecyclableMemoryStream:

```cs
using System;
using System.Text;

namespace RecyclableMemoryStreamReader
{
    using System.IO;

    using Microsoft.IO;

    class Program
    {
        static void Main(string[] args)
        {
            var memoryStreamManager = new RecyclableMemoryStreamManager();

            using (var inputFileMemoryStream = memoryStreamManager.GetStream())
            {
                var buffer = Encoding.Default.GetBytes("Some Text\nSome Other Text");
                inputFileMemoryStream.Write(buffer, 0, buffer.Length);
                inputFileMemoryStream.Position = 0L;

                using (var streamReader = new StreamReader(inputFileMemoryStream))
                {
                    string currentLine;
                    while ((currentLine = streamReader.ReadLine()) != null)
                    {
                        Console.WriteLine(currentLine);
                    }
                }
            } // << this throws the double-dispose / InvalidOperationException
        }
    }
}
```

doubleyewdee:
Yep, this is definitely wrong. Will patch it up. Thanks!

doubleyewdee:
Pending via https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream/pull/9

doubleyewdee:
Merged.

PR #9, opened 2015-03-23
Add missing calls to CheckDisposed in RecyclableMemoryStream. Integrate ...

doubleyewdee:
- Bring over code from pull request https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream/pull/4 (with modest changes).
- Fix a couple typos/spelling things.
- Do not throw on Dispose, from issue https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream/issues/5
- Add some missing calls to CheckDisposed.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -540,7 +540,7 @@ namespace Microsoft.IO
             }
 
             int blockSize = this.memoryManager.BlockSize;
-            long end = this.Position + count;
+            long end = (long)this.position + count;
             // Check for overflow
             if (end > MaxStreamLength)
             {
@@ -581,7 +581,7 @@ namespace Microsoft.IO
             {
                 Buffer.BlockCopy(buffer, offset, this.largeBuffer, this.position, count);
             }
-            this.Position = end;
+            this.position = (int)end;
             this.length = Math.Max(this.position, this.length);
         }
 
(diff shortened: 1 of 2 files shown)


src/RecyclableMemoryStream.cs
```diff
@@ -538,46 +539,42 @@ public override void Write(byte[] buffer, int offset, int count)
                 throw new ArgumentException("count must be greater than buffer.Length - offset");
             }
 
+            int blockSize = this.memoryManager.BlockSize;
+            long end = this.Position + count;
```
IDisposable: This should be ```this.position``` (note lower-case) to avoid redundant call to ```Check_Disposed()```

src/RecyclableMemoryStream.cs
```diff
@@ -538,46 +539,42 @@ public override void Write(byte[] buffer, int offset, int count)
...
-            int requiredBuffers = (end + blockSize - 1) / blockSize;
+            long requiredBuffers = (end + blockSize - 1) / blockSize;
             
             if (requiredBuffers * blockSize > MaxStreamLength)
             {
                 throw new IOException("Maximum capacity exceeded");
             }
 
-            EnsureCapacity(end);
+            this.EnsureCapacity((int)end);
```
IDisposable: This will now do a double-call to ```Check_Disposed()```

IDisposable:
Looks good, thanks for doing the patch up. Just one note, the call to ```this.Position``` instead of ```this.position``` will result in a double-call of ```this.Check_Position()```

doubleyewdee:
Thanks. Will update the code as mentioned and update the PR. Ben has been out this week, so I'm still waiting for him to review and merge.

PR #17, opened 2015-07-20
Additional assembly for net40

maxwellb:
Support a net40 build. Linked files for build, rewrote a blank Events class. Let me know if that would cause issues with serialization or anything - but it looked to be just internal.

### DIFF
--- a/Microsoft.IO.RecyclableMemoryStream.sln
+++ b/Microsoft.IO.RecyclableMemoryStream.sln
@@ -1,11 +1,15 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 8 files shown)


doubleyewdee:
Thanks Max,

I'm going to talk to Ben about pulling the EventSource stuff out (it needs work on our end though) since we can drive that entirely with events. We have a separate issue for supporting this as a PCL and that's the only thing stopping us at present.

That said, I'd love to encourage folks to move to .NET 4.5+ in general as well. :)

stakx:
> I'd love to encourage folks to move to .NET 4.5+ in general as well.

I think it would be great to achieve this by offering additional benefits instead of by force. Updating a large project to .NET 4.5+ because a (rather important) infrastructural component requires it often isn't feasible.

A better way (IMHO) might be to offer an alternative  .NET 4 (or even earlier) version that relies on `System.Diagnostics.TraceSource` instead of ETW, and then point out the disadvantages (e.g. slower performance). People would then be free to choose whether it is worth updating to .NET 4.5 in order to get the benefits.

doubleyewdee:
This is resolved by #29.

PR #24, opened 2016-05-03
Adding SafeRead and SafeReadByte methods. These take the position as …

orokanasaru:
…an extra parameter. They use the code of the previous Read and ReadByte methods, but with this.position replaced with the passed in position parameter. Existing Read and ReadByte methods replaced with calls to SafeRead and SafeReadByte by passing in this.position as the position parameter. These changes were made to enable the creation of thread-safe read wrappers.

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -25,6 +25,7 @@ namespace Microsoft.IO.UnitTests
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


doubleyewdee:
It would be nice to see new UTs for these methods. Am I missing something?

ltrzesniewski:
@doubleyewdee I need this feature, but it doesn't seem to be released yet, so I'll have to pull the code directly for now. Are there any plans to update the NuGet package?

PR #28, opened 2016-08-31
Cleaning up members in Dispose. This fixes the bug where it is possible to get RMS in a state where it is returning to the pool multiple references of the blocks it owns. What this causes is other RMS instances to grab the same references to memory from the manager and then start overwriting each other. We hit this issue while using RMS with wrappers allowing us to have multiple readers on a single RMS stream.

ariktg:


elifonalan:
tabs vs spaces...

doubleyewdee:
+1 on fixing tabs v. spaces (always spaces)

Not clear to me what bug this fixes?

ariktg:
This fixes the bug where it is possible to get RMS in a state where it is returning to the pool multiple references of the blocks it owns. What this causes is other RMS instances to grab the same references to memory from the manager and then start overwriting each other. We hit this issue while using RMS with wrappers allowing us to have multiple readers on a single RMS stream.

elifonalan:
Awesomesauce

PR #41, opened 2017-04-04
Add CBT support, require MSBuild 15, support .NET Standard 1.4

doubleyewdee:
Fixes:
- AssemblyVersion not in sync w/ NuGet package version.
- Adds use of CommonBuildToolset (https://github.io/commonbuildtoolset) for build extensions
- Migrate to MSBuild 15
- Target .NET Standard 1.4 in addition to 4.0, 4.5
- Some code cleanup required by migration

Stuff to look at (code-wise):
- Changed Dispose check in RecyclableMemoryStream to use long disposed member in order to switch to Interlocked.Read (required for netstandard14)
- Changed EventSource static writer to be named Writer (avoids overload confusion with new static Write methods in newer versions of EventSource e.g. netstandard14)
- Alternate (non-override) signature for netstandard14 Close and GetBuffer methods on RecyclableMemoryStream (they are not members of MemoryStream) -- the GetBuffer method is interesting, netstandard users will need to ensure they use RMS and not MS if they need to peel off large buffers for whatever reason.


arpit-nagar:
As we already added support for netStandard in master.
Can you please publish a nuget package which supports net standard.

brunoblank:
I tried the source today (compiled for netstandard 1.6) and TryGetBuffer did return false. 
Is the netstandard support complete?

PR #51, opened 2017-08-07
Adding custom StreamWriter for the RecyclableMemoryStream

aleksk:
The default StreamWriter allocates and uses two buffers: char[] to cache the input before feeding it to the encoder, and byte[] for the encoder to encode into.
Since we're already operating with the byte[], it doesn't make sense to have and use that additional byte[] only to make another memory copy to the backing stream afterwards.

Current CR proposes an implementation of such custom writer, which tries to encode directly to the stream's block/largeBuffer. It falls back to the legacy behavior (with temporary byte[]), if the write (flush) should span across multiple blocks.

---
Changes:
* Changed RecyclableMemoryStreamManager's GetStream() methods to return RecyclableMemoryStream instead of the MemoryStream. This will not break backcompat and will not require clients to cast to the public RecyclableMemoryStream.
* Added the internal GetCurrentBlockAndRelativeOffset() method to the RecyclableMemoryStream. It's used by the RecyclableMemoryStreamWriter.
* Added RecyclableMemoryStreamWriter that can write directly to the RecyclableMemoryStream, avoiding encoding to the temporary buffer first.
* Added unit tests.
* Removed trailing whitespaces.

---
Performance between the standard StreamWriter and RecyclableMemoryStreamWriter is pretty much the same, except the latter doesn't do that much memory copies.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -1,17 +1,17 @@
 ﻿// The MIT License (MIT)
-// 
+//
 // Copyright (c) 2015-2016 Microsoft
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 // FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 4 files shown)


src/RecyclableMemoryStreamWriter.cs
```diff
@@ -0,0 +1,561 @@
...
+        /// <param name="count">The maximum number of characters to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">The string length minus <paramref name="index"/> is less than <paramref name="count"/>.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="index"/> or <paramref name="count"/> is negative.</exception>
+        /// <exception cref="ObjectDisposedException">The current writer or the <see cref="BaseStream"/> is closed.</exception>
+        public virtual Task WriteSubstringAsync(string value, int index, int count)
+        {
+            var tuple = new Tuple<RecyclableMemoryStreamWriter, string, int, int>(this, value, index, count);
+#if NET40
```
aleksk: Wrap the entire method with the #if NET40?

src/RecyclableMemoryStreamWriter.cs
```diff
@@ -0,0 +1,561 @@
...
+            // blockBuffer can be null after this call, if the stream is currently full.
+            int blockBufferOffset;
+            byte[] blockBuffer = this.stream.GetCurrentBlockAndRelativeOffset(out blockBufferOffset);
+
+            int freeBytesInBlock = blockBuffer != null ? (blockBuffer.Length - blockBufferOffset) : 0;
+
+            if (freeBytesInBlock >= maxBytesToWrite)
+            {
+                // There is enough space in the current block, use it directly.
+                int bytesWritten = this.encoder.GetBytes(this.charBuffer, 0, this.charPos, blockBuffer, blockBufferOffset, flush: flushEncoder);
```
aleksk: Debug.Assert(blockBuffer != null) here?

src/RecyclableMemoryStream.cs
```diff
@@ -831,6 +830,46 @@ private BlockAndOffset GetBlockAndRelativeOffset(int offset)
...
+        /// Returns the buffer at the current stream position.
+        /// </summary>
+        /// <param name="blockOffset">
+        /// Gets assigned to the free space offset within the returned block.<br/>
+        /// It's set to <c>-1</c>, if there is no buffer to be returned.
+        /// </param>
+        /// <returns>
+        /// Buffer at the current position, or <c>null</c>, if the stream is full.
+        /// </returns>
+        internal byte[] GetCurrentBlockAndRelativeOffset(out int blockOffset)
```
aleksk: If this method is made protected, better return a struct/class here.

aleksk:
Will reopen with changes small enough to enable this scenario.

PR #52, opened 2017-08-18
Adding the public RecyclableMemoryStream::TryGetCurrentBlock method

aleksk:
This CR makes it possible to access the internal block currently being used by the RecyclableMemoryStream, so clients can write their own custom writers that write directly to the stream's block without calling the GetBuffer() method.


src/RecyclableMemoryStream.cs
```diff
@@ -475,6 +474,48 @@ public override byte[] ToArray()
...
+                var blockAndOffset = this.GetBlockAndRelativeOffset(this.position);
+
+                if (blockAndOffset.Block < this.blocks.Count)
+                {
+                    byte[] currentBlock = this.blocks[blockAndOffset.Block];
+                    block = new ArraySegment<byte>(currentBlock, 0, blockAndOffset.Offset);
+                }
+                else
+                {
+                    byte[] currentBlock = this.blocks[blockAndOffset.Block - 1];
```
aleksk: Need comment. This happens, if the stream is full.

src/RecyclableMemoryStream.cs
```diff
@@ -475,6 +474,48 @@ public override byte[] ToArray()
...
+        /// Returns information about the currently used <c>byte[]</c> block and the amount of data written to it.
+        /// </summary>
+        /// <param name="block">
+        /// Assigned to the <see cref="ArraySegment{byte}"/> that holds a reference to the current <c>byte[]</c>
+        /// block and data length in it (<see cref="ArraySegment{byte}.Count"/>).
+        /// </param>
+        /// <returns>
+        /// <c>true</c> if the block at the current position exists and was retrieved;
+        /// <c>false</c> if the current instance is disposed.
+        /// </returns>
```
aleksk: Need more details about the behavior when the stream is full, and when the large buffer is used.

doubleyewdee:
Sorry for sleeping on this -- I'm a little hesitant to expose internal implementation details (it would prevent any future change). We might just want to hop on a chat and discuss this for 30m?

aleksk:
Sure. Feel free to S+ or ping me over the IM.

benmwatson:
I agree with @doubleyewdee that this doesn't feel like an appropriate change for this library. I'm not sure if anything came of your conversation, but I'm inclined to close this.

Issue #48, opened 2017-07-12
Performance opportinities


buybackoff:
I have just played with a profiler and massively increased the performance of my RMS fork for my primary use case of write/read small buffers of several Kbs.  https://github.com/Spreads/Spreads/commit/cbeac8f6a295b007b1110dc31d1100f0a69cc08a

* Aggressively inlining all methods that are used internally. E.g. Capacity.
* Making all methods related to event tracing conditional to a compiler symbol to avoid needless method calls.
* Using System.Buffers array pool (modified to optionally return exact buffer sizes when requested) instead of ConcurrentBag.
* Using [vectorized memory copy](https://github.com/Spreads/Spreads/commit/a18b6360877c8583a75402584bd15269210ae09d) instead of `Buffer.BlockCopy`.
* Pooling of RMS instances using Roslyn's ObjectPool implementation.
* Manual implementation of `WriteByte` instead of redirecting to `Write` via a temporary `byte[1]`. I also added non-virtual SafeWriteByte.
* Using ThrowHelper for better inlining and less code size.

Some of this is directly applicable to the original implementation. My fork is now incompatible with upstream so I cannot create a PR, just put it there FYI & discussion.

I had the idea of integrating RMS with System.Buffers shared pool for a long time - not only this is faster, but also reduces memory by avoiding a separate pool. But the default shared pool implementation could return a buffer larger that the requested size, so I created a custom implementation with a parameter `exactSize`. Without shared pool modifications RMS could work with ArraySegments internally - that shouldn't be slower. And given that the shared pool returns a buffer that could be larger only by a power of two, that larger buffer could be split into several blocks and RMS will just increase capacity by more than one block. In `Dispose()` we just need to check if currently returning ArraySegment has the same buffer as previous and do not return it twice.

clarkis117:
@buybackoff  There's also an atomic dispose check being called before most methods. I'm not exactly sure that if that check needs to be atomic.

This:
```
       private bool Disposed => Interlocked.Read(ref this.disposedState) != 0;

        private void CheckDisposed()
        {

            if (this.Disposed)
            {

                throw new ObjectDisposedException($"The stream with Id {this.id} and Tag {this.tag} is disposed.");
            }
        }
```

buybackoff:
[...]

PR #56, opened 2017-11-20
Adding optimized manual implementation of WriteByte

clarkis117:
This replaces the implementation of the WriteByte method with a manual implementation, that doesn't forward calls to Write(byte[]) via a byte[1] field.

Optimization was identified in issue #48 

Benchmark: Writing 1MB using WriteByte (via Benchmark.NET, using x64 only)

|       Method |        Job | Runtime |     Toolchain |     Mean |     Error |    StdDev |
|------------- |----------- |-------- |-------------- |---------:|----------:|----------:|
| WriteByteNew |      Net47 |     Clr |   CsProjnet47 | 24.89 ms | 0.0104 ms | 0.0097 ms |
| WriteByteOld |      Net47 |     Clr |   CsProjnet47 | 56.09 ms | 0.0703 ms | 0.0623 ms |
| WriteByteNew | NetCore1.1 |    Core | .NET Core 1.1 | 31.25 ms | 0.0448 ms | 0.0419 ms |
| WriteByteOld | NetCore1.1 |    Core | .NET Core 1.1 | 61.21 ms | 0.0311 ms | 0.0276 ms |
| WriteByteNew | NetCore2.0 |    Core | .NET Core 2.0 | 21.31 ms | 0.0039 ms | 0.0037 ms |
| WriteByteOld | NetCore2.0 |    Core | .NET Core 2.0 | 50.01 ms | 0.0243 ms | 0.0215 ms |

```
BenchmarkDotNet=v0.10.12, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.214)
Intel Core i7-6800K CPU 3.40GHz (Skylake), 1 CPU, 12 logical cores and 6 physical cores
Frequency=3326080 Hz, Resolution=300.6542 ns, Timer=TSC
.NET Core SDK=2.1.300-preview2-008065
  [Host]     : .NET Core 2.0.5 (Framework 4.6.26020.03), 64bit RyuJIT
  Net47      : .NET Framework 4.7 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0
  NetCore1.1 : .NET Core 1.1.5 (Framework 4.6.25815.04), 64bit RyuJIT
  NetCore2.0 : .NET Core 2.0.5 (Framework 4.6.26020.03), 64bit RyuJIT

Jit=RyuJit  Platform=X64
```

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -68,12 +68,6 @@ namespace Microsoft.IO
         /// </summary>
         private readonly List<byte[]> blocks = new List<byte[]>(1);
 
-        /// <summary>
-        /// This buffer exists so that WriteByte can forward all of its calls to Write
-        /// without creating a new byte[] buffer on every call.
-        /// </summary>
-        private readonly byte[] byteBuffer = new byte[1];
-
         private readonly Guid id;
 
         private readonly RecyclableMemoryStreamManager memoryManager;
@@ -626,10 +620,33 @@ namespace Microsoft.IO
         /// <exception cref="ObjectDisposedException">Object has been disposed</exception>
         public override void WriteByte(byte value)
         {
-            this.CheckDisposed();
-            this.byteBuffer[0] = value;
-            this.Write(this.byteBuffer, 0, 1);
-        }
+			this.CheckDisposed();
+			int blockSize = this.memoryManager.BlockSize;
+			int end = this.position + 1;
+			// Check for overflow
+			if (end > MaxStreamLength)
+			{
+				throw new IOException("Maximum capacity exceeded");
+			}
+
+			this.EnsureCapacity(end);
+
+			if (this.largeBuffer is null)
+			{
+				int block = this.position / blockSize, offset = this.position % blockSize;
+
+				byte[] currentBlock = this.blocks[block];
+
+				currentBlock[offset] = value;
+			}
+			else
+			{
+				this.largeBuffer[this.position] = value;
+			}
+
+			this.position = end;
+			this.length = Math.Max(this.position, this.length);
+		}
 
         /// <summary>
         /// Reads a single byte from the current position in the stream.
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -620,33 +620,33 @@ namespace Microsoft.IO
         /// <exception cref="ObjectDisposedException">Object has been disposed</exception>
         public override void WriteByte(byte value)
         {
-			this.CheckDisposed();
-			int blockSize = this.memoryManager.BlockSize;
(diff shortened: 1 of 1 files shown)


src/RecyclableMemoryStream.cs
```diff
@@ -626,10 +620,33 @@ public override string ToString()
...
+			{
+				throw new IOException("Maximum capacity exceeded");
+			}
+
+			this.EnsureCapacity(end);
+
+			if (this.largeBuffer is null)
+			{
+				int block = this.position / blockSize, offset = this.position % blockSize;
+
```
buybackoff: `offset = this.position - block * blockSize` will save "expensive" modulo operation, you already have the division result.

clarkis117:
@doubleyewdee is the project dead?

shatl:
guys, any updates on this ?

clarkis117:
@doubleyewdee I poked, who I'm guessing is the maintainer on twitter a few weeks ago, this is the response I received: https://twitter.com/doubleyewdee/status/951285541319843840
Although, I should probably update the benchmark to include results for .NET Core 1.x, 2.0, and 2.1; now that I got Benchmark.NET working with them on my dev box 👍

benmwatson:
I am finally getting around to reviewing this... I'm curious what the motivation for this is. When I originally wrote this, I think I assumed that WriteByte would be a rare call (because who writes to a stream one byte at a time?), thus preferred avoiding code duplication.

That said, I'm not opposed to this change.

raymondpwilson:
I just discovered this branch and PR after independently optimising the WriteByte() method (in addition to a collection of other optimisations).

I have included my version of this function below if you are interested.

    public override void WriteByte(byte value)
    {
      this.CheckDisposed(); //<-- Editorial: The run time for this dominates WriteByte() performance. IFDEFing out this call (assuming your environment is safe to do so) markedly increases performance of this method.

      long end = (long)this.position + 1;

      // Check for overflow
      if (end > MaxStreamLength)
      {
        throw new IOException("Maximum capacity exceeded");
      }

      if (end >= this.capacity)  // <-- Note: Another optimisation that uses a cached capacity via backing variable rather than using the expensive Capacity property.
      {
        this.EnsureCapacity((int) end); // <- EnsureCapacity() is expensive - only call it if you need to
      }

      if (this.largeBuffer == null)
      {
        this.blocks[this.position / this.blockSize][this.position % this.blockSize] = value;
      }
      else
      {
        this.largeBuffer[this.position] = value;
      }

      this.position = (int)end;
      this.length = Math.Max(this.position, this.length);
    }

benmwatson:
I'm not convinced of proper use cases for this. Writing 1MB stream one byte at a time feels like an artificial benchmark. I would prefer cleaner code right now.

PR #59, opened 2018-06-05
Convert Enumerable.Sum to a for[each]

ndrwrbgs:
The compiler can optimize a for loop over an array very well, but cannot optimize LINQ (and LINQ does not have overloads to special handle arrays today). This results in CallVirt methods that cannot be inlined and add additional performance cost (this can be seen in AzProfiler logs internally).


danielmarbach:
@doubleyewdee this seems like a reasonable thing to merge and release. Any ETAs when it will be worthwhile ;) ?

Issue #60, opened 2018-07-02
Guid generation is expensive


nospof:
Hello

I've recently started using this library to optmise a hot path in order to reduce allocations but I've found that creating a new RecyclableMemoryStream is quite expensive:

![image](https://user-images.githubusercontent.com/816807/42167299-86f0ad9a-7e05-11e8-84a3-bd15108b0f16.png)

The id field (Guid) on the RecyclableMemoryStream is being initialized on the constructor regardless of whether ETW tracing is enabled or not and generating a new Guid is expensive (interop call + 16 bytes allocated).

I've changed the constructor code to only initialize the field if ETW tracing is enabled and it now looks like this:

![image](https://user-images.githubusercontent.com/816807/42167353-bd4102b4-7e05-11e8-8123-2610e9b74a63.png)

I've pushed my changes in case you find this solution sensible.

PR #61, opened 2018-07-02
Avoid allocation/generation of Guid in RecyclableMemoryStream if ETW tracing is disabled

nospof:
Addresses #60 

Only initializes the id field of the RecyclableMemoryStream if ETW tracing is enabled (id is only used for event correlation).

This means that only RecyclableMemoryStreams allocated after ETW is enabled will have a correlation id generated, otherwise it will be Guid.Empty.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -187,8 +187,12 @@ namespace Microsoft.IO
             : base(emptyArray)
         {
             this.memoryManager = memoryManager;
-            this.id = Guid.NewGuid();
             this.tag = tag;
+
+            if (RecyclableMemoryStreamManager.Events.Writer.IsEnabled())
+            {
+                this.id = Guid.NewGuid();
+            }
 
             if (requestedSize < memoryManager.BlockSize)
             {


danielmarbach:
@doubleyewdee this seems like a reasonable thing to merge and release. Any ETAs when it will be worthwhile ;) ?

benmwatson:
>(id is only used for event correlation).

I disagree with this assertion. Guids are also useful for debugging--they are the only way to distinguish streams from each other.

What is the reason for this change? Guids are structs, so there is no GC alloc.

nospof:
@benmwatson the main reason for this change is to speed up creating new instances of RecyclableMemoryStream, which I used for serializing/deserializing messages in a financial trading backend.
My focus is not really to reduce GC allocs but to make sure we don't make a relatively expensive call if we can avoid it. In the end I could not use RecyclableMemoryStream to serialize/deserialize small message payloads because the cost of creating the Guid was prohibitive in the hot path. Instead I only use it for infrequent but larger message payloads.

I agree that this would in practice reduce your ability to easily debug and track allocations of RecyclableMemoryStream instances since the id field would be empty in a debugging session. However, you could argue that enabling ETW tracing is good enough to diagnose issues? Additionally, for devs using VS, the ability to activate debugger ObjectIDs is good enough?

btw, love your book! 👍

benmwatson:
I'm uncomfortable with the implicit nature of this--just because ETW is disabled doesn't feel like a strong enough reason. What if there were an overload of the ctor that took an existing Guid? Then you could pass Guid.Empty. That gives you control over the performance without the implicit behavior.

nospof:
@benmwatson that makes sense. I'll push another PR with those changes.

PR #63, opened 2018-07-11
Add two-stage allocation

magicpanda0618:
Call GetStream to get a MemoryStream like before, and call GetTwoStageStream to get a MemoryStream which allows TinyBlock allocation.


benmwatson:
Rejecting this after talking in person. We have a better strategy.

Issue #66, opened 2018-11-12
override Stream methods that accept Span and Memory


tmds:
This avoids allocations and copies in the Stream base class.

benmwatson:
See @brantburnett 's changes. If there is more you wish done, please open a new issue, or submit a PR.

PR #68, opened 2019-04-05
Implement Span<byte>/ReadOnlySpan<byte> overloads

brantburnett:
Addresses #66

.NET Core 2.1 implements Read and Write overloads that accept Span
constructs. These constructs will work without overloading, but the
default implementations work via inefficient byte array copies.

This adds a netcoreapp2.1 target to the NuGet package, and implements
efficient overloads of Read(Span<byte>) and Write(ReadOnlySpan<byte>).
Overloads of ReadAsync and WriteAsync are not required, they have
efficient default implementations that will use these methods as needed.

The unit tests have also been updated to target .NET Core 2.1 so that
they can test the new method implementations.


shatl:
Hi, could you add NetStandard 2.1 as target?

brantburnett:
@shatl 

My understanding was the .NET Standard 2.1 is still in preview and not finalized yet. I was assuming we'd add it later once it was finalized.

I just attempted to add it, using VS2019, and I get errors about the .NET SDK not supporting targetting .NET Standard 2.1. Is there some way to pull the newer SDK preview that supports this?

Numpsy:
> Is there some way to pull the newer SDK preview that supports this?

You can install the most recent preview of the .NET Core 3.0 SDK to get the option to target Standard 2.1.

shatl:
Hi,
You can use
https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script to
install 3.0 preview

brantburnett:
@shatl @Numpsy 

Thanks for the pointers, I've pushed another commit which also targets .NET Standard 2.1. I left the netcoreapp2.1 target in place so that the methods will still be available in .NET Core 2.1 as well, since Core 2.1 uses Standard 2.0.

brantburnett:
@benmwatson @doubleyewdee Not sure if you guys are the right people to ping about this PR? I noticed you merged some of the commits in the past. If not, do you know who maintains this now?

benmwatson:
@brantburnett Yes we're the right people. We've had a very hands-off approach for quite a while, but I would like to change that. Your change is a bit more involved than the typical ones, so let me take a careful look later this week.

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "3.0.100-preview3-010431"
```
benmwatson: Update to latest official preview: 3.0.100-preview5-011568

brantburnett: Done

cocowalla:
Should there also be an overload of `RecyclableMemoryStreamManager.GetStream` that takes a `Memory<byte>`?

brantburnett:
@cocowalla makes sense to me

cocowalla:
@brantburnett OK, I'll open a new issue for it

PR #82, opened 2020-02-03
Update documentation

benmwatson:
Added significant documentation updates in README.md. Also added a diagram and updated the CHANGES.md file with last release.


README.md
```diff
@@ -21,14 +32,9 @@ Install-Package Microsoft.IO.RecyclableMemoryStream
...
-For more details, refer to the [announcement blog post](http://www.philosophicalgeek.com/2015/02/06/announcing-microsoft-io-recycablememorystream/)
-
 ## Build Targets
 
-The code ships utilizing CommonBuildToolset. This provides some build extensions over top of MSBuild to automatically pull
-NuGet packages from command line build, etc.
-
-MSBuild 15 is required to build the code. You get this with Visual Studio 2017. Cross-platform builds will be documented
+At least MSBuild 15 is required to build the code. You get this with Visual Studio 2017.
 in a future release.
```
erikness: straggler line

README.md
```diff
@@ -37,4 +43,87 @@ Build targets are: net40, net45, netstandard1.4, netstandard2.1, and netcoreapp2
...
+- Be sure to set the default processor architecture for tests to x64 (or the giant allocation test will fail)
+
+## How It Works
+
+RecyclableMemoryStream improves GC performance by ensuring that the larger buffers used for the streams are put into the gen 2 heap and stay there forever. This should cause full collections to happen less frequently. If you pick buffer sizes above 85K then you will ensure these are placed on the large object heap, which is touched even less frequently by the garbage collector.
+
+The MemoryStreamManager class maintains two separate pools of objects:
+
+1. **Small Pool** - Holds small buffers (of configurable size). Used by default for all normal read/write operations. Multiple small buffers are chained together in the RecyclableMemoryStream class and abstracted into a single stream.
+2. ** Large Pool** - Holds large buffers, which are only used when you must have a single, contiguous buffer, such as a call to GetBuffer().
```
erikness: straggler space after the first "**"

README.md
```diff
@@ -37,4 +43,87 @@ Build targets are: net40, net45, netstandard1.4, netstandard2.1, and netcoreapp2
...
+## How It Works
+
+RecyclableMemoryStream improves GC performance by ensuring that the larger buffers used for the streams are put into the gen 2 heap and stay there forever. This should cause full collections to happen less frequently. If you pick buffer sizes above 85K then you will ensure these are placed on the large object heap, which is touched even less frequently by the garbage collector.
+
+The MemoryStreamManager class maintains two separate pools of objects:
+
+1. **Small Pool** - Holds small buffers (of configurable size). Used by default for all normal read/write operations. Multiple small buffers are chained together in the RecyclableMemoryStream class and abstracted into a single stream.
+2. ** Large Pool** - Holds large buffers, which are only used when you must have a single, contiguous buffer, such as a call to GetBuffer(). 
+
+A RecyclableMemoryStream starts out by using a small buffer, chaining additional ones as the stream capacity grows. Should you ever call `GetBuffer()` and the length is greater than single small buffer's capacity, then the small buffers are converted to a single large buffer.
```
erikness: maybe worth mentioning that you can set a stream size from the start? i.e.

"A RecyclableMemoryStream starts out by using a small buffer, chaining additional ones as the stream capacity grows **(you can also set a start capacity and the stream will have potentially multiple small buffers at the start)**"

your call - this might be too obvious, or perhaps too much detail too soon.

erikness: Worth mentioning here: when a stream is Disposed (or goes out of scope? I forget), its buffers are released into the pool (and zeroed out? I also forget this) instead of being eligible for garbage collection.

README.md
```diff
@@ -37,4 +43,87 @@ Build targets are: net40, net45, netstandard1.4, netstandard2.1, and netcoreapp2
...
+
+## How It Works
+
+RecyclableMemoryStream improves GC performance by ensuring that the larger buffers used for the streams are put into the gen 2 heap and stay there forever. This should cause full collections to happen less frequently. If you pick buffer sizes above 85K then you will ensure these are placed on the large object heap, which is touched even less frequently by the garbage collector.
+
+The MemoryStreamManager class maintains two separate pools of objects:
+
+1. **Small Pool** - Holds small buffers (of configurable size). Used by default for all normal read/write operations. Multiple small buffers are chained together in the RecyclableMemoryStream class and abstracted into a single stream.
+2. ** Large Pool** - Holds large buffers, which are only used when you must have a single, contiguous buffer, such as a call to GetBuffer(). 
+
```
erikness: Perhaps worth mentioning at this point: buffers above the max large pool size (i.e. the ones that fire the event MemoryStreamNonPooledLargeBufferCreated) will simply be allocated as if it were a normal MemoryStream, and the buffer will be eligible for garbage collection when it is disposed / goes out of scope.

One of the nice implications of this design is that if you have to allocate a very large stream, it doesn't steal buffers from the bounded pools. (I helped someone with this problem once, where they were concerned that the occasional huge RMS, like 100+MB, would starve the RMS manager).

API.md
```diff
@@ -0,0 +1,337 @@
...
+## Properties
+
+| Name | Type | Summary |
+|---|---|---|
+| **Id** | Guid | Unique identifier for this stream across it's entire lifetime |
+| **Tag** | string | A temporary identifier for the current usage of this stream. |
+| **MemoryManager** | [RecyclableMemoryStreamManager](#recyclablememorystreammanager-class) | Gets the memory manager being used by this stream. |
+| **AllocationStack** | string | Callstack of the constructor. It is only set if MemoryManager.GenerateCallStacks is true,<br>which should only be in debugging situations. |
+| **DisposeStack** | string | Callstack of the Dispose call. It is only set if MemoryManager.GenerateCallStacks is true,<br>which should only be in debugging situations. |
+| **Capacity** | int | Gets or sets the capacity |
```
nycdotnet: how can capacity be an int and length a long?  Also why would I ever set this?  what happens if I write past the capacity?

API.md
```diff
@@ -0,0 +1,337 @@
+# Microsoft.IO.RecyclableMemoryStream.dll v.1.3.2.0 API documentation
+
+Created by 
+[mddox](https://github.com/loxsmoke/mddox) on 2/3/2020
+
+# All types
```
nycdotnet: move the types that have only auto generated docs to the bottom.  If that will cause an issue with future generation refreshes, consider a quick index here of the most interesting types (which are the only ones people will actually look at)

API.md
```diff
@@ -0,0 +1,337 @@
...
+The stream is implemented on top of a series of uniformly-sized blocks. As the stream's length grows,
+additional blocks are retrieved from the memory manager. It is these blocks that are pooled, not the stream
+object itself.
+
+The biggest wrinkle in this implementation is when GetBuffer() is called. This requires a single 
+contiguous buffer. If only a single block is in use, then that block is returned. If multiple blocks 
+are in use, we retrieve a larger buffer from the memory manager. These large buffers are also pooled, 
+split by size--they are multiples/exponentials of a chunk size (1 MB by default).
+
+Once a large buffer is assigned to the stream the blocks are NEVER again used for this stream. All operations take place on the
```
nycdotnet: the small blocks?

API.md
```diff
@@ -0,0 +1,337 @@
...
+
+Once a large buffer is assigned to the stream the blocks are NEVER again used for this stream. All operations take place on the 
+large buffer. The large buffer can be replaced by a larger buffer from the pool as needed. All blocks and large buffers 
+are maintained in the stream until the stream is disposed (unless AggressiveBufferReturn is enabled in the stream manager).
+
+## Properties
+
+| Name | Type | Summary |
+|---|---|---|
+| **Id** | Guid | Unique identifier for this stream across it's entire lifetime |
```
nycdotnet: its

API.md
```diff
@@ -0,0 +1,337 @@
...
+|---|---|---|
+| **Close()** | void | Equivalent to Dispose |
+| **GetBuffer()** | byte[] | Returns a single buffer containing the contents of the stream.<br>The buffer may be longer than the stream length. |
+| **Read(byte[] buffer, int offset, int count)** | int | Reads from the current position into the provided buffer |
+| **ReadByte()** | int | Reads a single byte from the current position in the stream. |
+| **SafeRead(byte[] buffer, int offset, int count, Int32& streamPosition)** | int | Reads from the specified position into the provided buffer |
+| **SafeReadByte(Int32& streamPosition)** | int | Reads a single byte from the specified position in the stream. |
+| **Seek(long offset, SeekOrigin loc)** | long | Sets the position to the offset from the seek location |
+| **SetLength(long value)** | void | Sets the length of the stream |
+| **ToArray()** | byte[] | Returns a new array with a copy of the buffer's contents. You should almost certainly be using GetBuffer combined with the Length to <br>access the bytes in this stream. Calling ToArray will destroy the benefits of pooled buffers, but it is included<br>for the sake of completeness. |
```
nycdotnet: I have seen benefits from using this library to build up a long “stream” from lots of small chunks, but the library I am required to pass the output to only takes a byte[].  Is ToArray my best option?  Could I do some sort of span thing? Any chance this library could integrate with renting arrays?

README.md
nycdotnet: not too obvious- this stuff is great

README.md
```diff
@@ -37,4 +43,186 @@ Build targets are: net40, net45, netstandard1.4, netstandard2.1, and netcoreapp2
...
+* `GetBuffer` - If possible, a reference to the single block will be returned to the caller. If multiple blocks are in use, they will be converted into a single large pool buffer and the data copied into it. In all cases, the caller must use the `Length` property to determine who much usable data is actually in the returned buffer. If the stream length is longer than the maximum allowable stream size, a single buffer will still be returned, but it will not be pooled.
+* `ToArray` - It looks similar to `GetBuffer` on the surface, but is actually significantly different. In `ToArray` the data is *always* copied into a new array that is exactly the right length for the full contents of the stream. This new buffer is never pooled. Users of this library should consider any call to `ToArray` to be a bug, as it wipes out many of the benefits of `RecyclableMemoryStream` completely. We may even add a feature
```
nycdotnet: ... ?

README.md
benmwatson: I mention the disposal mechanism in a later paragraph in this section.

Buffers are not zeroed.

API.md
benmwatson: Unfortunately, the types are determined by the base class, MemoryStream, and I'm adhering to those. I think it makes no sense, but it is what it is.

I'll add some documentation about exceeding the capacity somewhere.

API.md
benmwatson: If the API only takes a byte[] with no length parameter, you might be stuck using ToArray, which stinks, because you lose all the pooling benefits. I'm not sure there is another option--it all comes down to the APIs you have to call, even with Span.

API.md
benmwatson: Yeah...I've been trying to find a good XML->MD translator and they mostly leave a lot to be desired. Trying a new one now that should resolve this.

README.md
```diff
@@ -21,20 +32,196 @@ Install-Package Microsoft.IO.RecyclableMemoryStream
...
+
+If you forget to call a stream's `Dispose` method, this could cause a memory leak. To help you prevent this, each stream has a finalizer that will be called by the CLR once there are no more references to the stream. This finalizer will raise an event or log a message about the leaked stream.
+
+## Usage
+
+You can jump right in with no fuss by just doing a simple replacement of MemoryStream with something like this:
+
+```
+var sourceBuffer = new byte[]{0,1,2,3,4,5,6,7};
```
nycdotnet: You say on line 90 that it should be declared once and live for the entire process - but I suspect there is some percentage of people who will not understand this.

imo the no-fuss jump-in is something like this (please rename as needed):

```csharp
    public static class Streams
    {
        public static readonly RecyclableMemoryStreamManager Manager = new RecyclableMemoryStreamManager();
    }

    public class MyClass
    {
        public void WriteToStream()
        {
            var sourceBuffer = new byte[] { 0, 1, 2, 3, 4, 5, 6, 7 };
            using (var stream = Streams.Manager.GetStream())
            {
                stream.Write(sourceBuffer, 0, sourceBuffer.Length);
            }
        }
    }
```

README.md
```diff
@@ -21,20 +32,196 @@ Install-Package Microsoft.IO.RecyclableMemoryStream
...
+To facilitate easier debugging, you can optionally provide a string tag, which serves as a human-readable identifier for the stream. In practice, I’ve usually used something like “ClassName.MethodName” for this, but it can be whatever you want. Each stream also has a GUID to provide absolute identity if needed, but the tag is usually sufficient.
+
+```
+using (var stream = manager.GetStream("Program.Main"))
+{
+    stream.Write(sourceBuffer, 0, sourceBuffer.Length);
+}
+```
+
+You can also provide an existing buffer. It’s important to note that this buffer will be copied into the pooled buffer:
```
nycdotnet:
```suggestion
You can also provide an existing buffer. It’s important to note that this buffer will be copied into the pooled buffers owned by RecyclableMemoryStream:
```

README.md
```diff
@@ -21,20 +32,196 @@ Install-Package Microsoft.IO.RecyclableMemoryStream
...
+                                                largeBufferMultiple, 
+                                                maxBufferSize);
+
+manager.GenerateCallStacks = true;
+manager.AggressiveBufferReturn = true;
+manager.MaximumFreeLargePoolBytes = maxBufferSize * 4;
+manager.MaximumFreeSmallPoolBytes = 100 * blockSize;
+```
+
+While this library strives to be very generic and not impose too many restraints on how you use it, its purpose is to reduce the cost of garbage collections incurred by frequent large allocations. Thus, there are some general guidelines for usage that may be useful to you:
```
nycdotnet:
```suggestion
While this library strives to be very general and not impose too many restraints on how you use it, its purpose is to reduce the cost of garbage collections incurred by frequent large allocations. Thus, there are some general guidelines for usage that may be useful to you:
```

nycdotnet:
imo the docs should explicitly state whether memory in the pooled buffers will be initialized or not - if not, calling GetBuffer() is a security risk due to the possibility the buffer will be bigger than the actual content.

nycdotnet:
This looks really great so far - thanks @benmwatson

README.md
benmwatson: Good point. Let me redo some of these examples.

PR #89, opened 2020-03-18
Add setting to throw an exception on ToArray calls

### LINKED ISSUE #81: Add a setting to disallow ToArray
Calling `ToArray` on a RecyclableMemoryStream should be considered a bug because it wipes out all the benefits of using the library. While the method does work as intended, we could add a setting on RecyclableMemoryStreamManager to cause an exception to be thrown when called.

benmwatson:
A new setting on RecyclableMemoryStreamManager will cause an exception to be thrown on any calls to ToArray.

Closes #81


src/RecyclableMemoryStream.cs
```diff
@@ -506,16 +506,24 @@ public override bool TryGetBuffer(out ArraySegment<byte> buffer)
...
         /// <exception cref="ObjectDisposedException">Object has been disposed</exception>
+        /// <exception cref="NotSupportedException">The current RecyclableStreamManager object disallows ToArray calls.</exception>
 #pragma warning disable CS0809
         [Obsolete("This method has degraded performance vs. GetBuffer and should be avoided.")]
         public override byte[] ToArray()
         {
             this.CheckDisposed();
+            
+            string stack = this.memoryManager.GenerateCallStacks ? Environment.StackTrace : null;
+            RecyclableMemoryStreamManager.Events.Writer.MemoryStreamToArray(this.id, this.tag, stack, this.length);
```
erikness: nice catch putting this.length into the event instead of just 0

PR #100, opened 2020-09-03
Expose internal buffers through GetReadOnlySequence()

grbell-ms:
# Problem #
Using RecyclableMemoryStream with other common APIs often requires unnecessary intermediary buffers. Consider something like IncrementalHash.AppendData(). There are two options to hash the contents of the stream:

### Use TryGetBuffer() ###
```csharp
IncrementalHash hash;
RecyclableMemoryStream stream;
// ...
stream.TryGetBuffer(out var buffer);
hash.AppendData(buffer);
```

or 

### Use your own buffer ###
```csharp
IncrementalHash hash;
RecyclableMemoryStream stream;
byte[] buffer;
int count;
// ...
while((count = stream.Read(buffer)) > 0)
    hash.AppendData(buffer, 0, count);
```

While sometimes TryGetBuffer() won't copy the stream contents to a new buffer, it will if the stream has multiple blocks. This costs CPU time, cache evictions, and increased working set.

# Solution #

GetReadOnlySequence() lets us work with APIs accepting arrays or spans without copying.

### Use GetReadOnlySequence() ###
```csharp
IncrementalHash hash;
RecyclableMemoryStream stream;
// ...
foreach(Memory<byte> memory in stream.GetReadOnlySequence())
    hash.AppendData(memory.Span);
```

There are many other uses as well: SequenceReader, Encoder.GetBytes(), reading from a network or file stream, and so on. We can improve performance in all these cases by exposing the stream's buffers directly through GetReadOnlySequence().

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -23,8 +23,10 @@
 namespace Microsoft.IO.UnitTests
 {
     using System;
+    using System.Buffers;
     using System.Collections.Generic;
     using System.IO;
+    using System.Runtime.InteropServices;
     using System.Threading.Tasks;
     
     using Microsoft.IO;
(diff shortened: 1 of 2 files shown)


benmwatson (approved):
Very nice

Issue #57, opened 2018-04-10
RecyclableMemoryStream class should override CopyToAsync


joshdean2000:
Because RecyclableMemoryStream is not overriding the CopyToAsync it means it's using the default one in Stream class which generates a byte[] buffer each time it's called.

This kind of defeats the purpose of using RecyclableMemoryStream.

Not sure why Stream doesn't support overriding CopyTo as well as this would also benefit from being overridden.

benmwatson:
The default implementation of Stream.CopyToAsync now uses ArrayPool and Memory for its temporary buffers. I don't think there is a strong reason to override this behavior at the moment.

brunom:
Overriding CopyTo* would avoid an unnecessary memcpy.

benmwatson:
Fair point. I'll try to look at this issue soon.

benmwatson:
@brunom @joshdean2000 Would you like to review the referenced PR for this issue?

paulomorgado:
Using `Task.Run` is async over sync and that's a bad practice. See [Should I expose asynchronous wrappers for synchronous methods?](https://devblogs.microsoft.com/pfxteam/should-i-expose-asynchronous-wrappers-for-synchronous-methods/).

The thread pool might be a critical component of the system using the stream.

But avoiding allocation is always good. And, if nothing else in the system is using `ArrayPool` there's no reason for `RecyclableMemoryStream` to use it just for this.

PR #107, opened 2020-11-03
Override CopyToAsync

benmwatson:
This resolves Issue #57 by implementing the `CopyToAsync` method.

If the target stream is a `MemoryStream`-derived type, then it will call `Write` synchronously immediately. Otherwise, it will call `WriteAsync` on the destination stream.


paulomorgado:
Using `Task.Run` is async over sync and that's a bad practice. See [Should I expose asynchronous wrappers for synchronous methods?](https://devblogs.microsoft.com/pfxteam/should-i-expose-asynchronous-wrappers-for-synchronous-methods/).

The thread pool might be a critical component of the system using the stream.

But avoiding allocation is always good.

benmwatson:
`Task.Run` is only called when the target stream is unknown, and an async write is done. In cases where a sync write can be done, a completed task is returned.

paulomorgado:
Nevertheless, this is a major no-no.

Besides that, what's the benefit of having `Task.Run` here?

PR #109, opened 2020-11-16
Support for .NET 5

jochenjonc:
Added net5.0 as TargetFramework

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.302",
+    "version": "5.0.100",
     "rollForward": "latestMinor"
   }
 }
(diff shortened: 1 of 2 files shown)


benmwatson:
I would rather not add a new target unless it is taking advantage of features present in that version. There is nothing to be gained by adding net5.0 at this time. 

I could see changing the rollForward value in global.json to build against later frameworks if you don't have 3.x installed.

jochenjonc:
Ok, I understand. But then maybe netcoreapp2.1 should also be removed.

DenSmoke:
I think adding net5 TFM is good for future improvements. I mean you can't add features from net5 without targeting to it.

benmwatson:
@DenSmoke If we add a feature that relies on net5, then we'll definitely need to add it then, but now, it's just not accurate. There are not net5-specific features.

@jochenjonc Both netstandard2.1 and netcoreapp2.1 are needed because they are targeting different things. netstandard2.1 is implemented by .NET Core 3.0, but there are features that .NET Core 2.1 supports, which aren't in netstandard2.0, hence the seeming duplication there. It's unfortunate, and a little messy.

I am actually considering dropping support for net40 soon, but it depends on how many more things are implemented that won't be supported back that far. I'd like the library to remain very flexible and able to be used under very different situations.

benmwatson:
Closing this because the library supports net5 as-is and there are no net5-specific methods yet.

PR #113, opened 2020-11-25
Remove Task.Run

paulomorgado:
This PR removes `Task.Run` from `CopyToAsync`.

To keep form creating the async state machine, only if more than one block exists, the async state machine will be created. This happens with the known [costs and benefits of eliding async](https://blog.stephencleary.com/2016/12/eliding-async-await.html).

Add more tests for more coverage.

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -2505,6 +2505,32 @@ namespace Microsoft.IO.UnitTests
         }
 
         [Test]
+        public void CopyToAsyncZeroBlocks()
+        {
+            using (var stream = GetDefaultStream())
+            {
+                var otherStream = GetDefaultStream();
+                stream.CopyToAsync(otherStream);
+                Assert.That(otherStream.Length, Is.EqualTo(0));
+            }
+        }
+
+        [Test]
+        public void CopyToAsyncZeroBlocksNonMemoryStream()
+        {
+            using (var stream = GetDefaultStream())
+            {
+                var filename = Path.GetRandomFileName();
+                using (var fileStream = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite, DefaultBlockSize, FileOptions.Asynchronous))
+                {
+                    stream.CopyToAsync(fileStream).Wait();
+                }
+                var otherBuffer = File.ReadAllBytes(filename);
+                Assert.That(otherBuffer.Length, Is.EqualTo(0));
+            }
+        }
+
+        [Test]
         public void CopyToAsyncOneBlock()
         {
             using (var stream = GetDefaultStream())
@@ -2515,6 +2541,24 @@ namespace Microsoft.IO.UnitTests
                 stream.CopyToAsync(otherStream);
                 Assert.That(otherStream.Length, Is.EqualTo(stream.Length));
                 RMSAssert.BuffersAreEqual(stream.GetBuffer(), otherStream.GetBuffer(), buffer.Length);
+            }
+        }
+
+        [Test]
+        public void CopyToAsyncOneBlockNonMemoryStream()
+        {
+            using (var stream = GetDefaultStream())
+            {
+                var buffer = GetRandomBuffer(DefaultBlockSize);
+                stream.Write(buffer, 0, buffer.Length);
+                var filename = Path.GetRandomFileName();
+                using (var fileStream = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite, DefaultBlockSize, FileOptions.Asynchronous))
+                {
+                    stream.CopyToAsync(fileStream).Wait();
+                }
+                var otherBuffer = File.ReadAllBytes(filename);
+                Assert.That(otherBuffer.Length, Is.EqualTo(stream.Length));
+                RMSAssert.BuffersAreEqual(stream.GetBuffer(), otherBuffer, buffer.Length);
             }
         }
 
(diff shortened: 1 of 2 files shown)


UnitTests/Tests.cs
```diff
@@ -2480,7 +2480,7 @@ public void CopyToAsyncThrowsOnNullDestination()
         }
 
         [Test]
-        public void CopyToAsyncThrowsIfDisposed()
+        public void CopyToAsyncThrowsIfDaksisposed()
```
benmwatson, line 2483: typo

src/RecyclableMemoryStream.cs
```diff
@@ -978,17 +995,17 @@ public override long Seek(long offset, SeekOrigin loc)
...
-                break;
-            case SeekOrigin.Current:
-                newPosition = (int)offset + this.position;
-                break;
-            case SeekOrigin.End:
-                newPosition = (int)offset + this.length;
-                break;
-            default:
-                throw new ArgumentException("Invalid seek origin", nameof(loc));
+                case SeekOrigin.Begin:
```
benmwatson, line 998: Looks like your editor is making some different spacing decisions than were in here previously. Can you revert all these lines that aren't part of the main change?

src/RecyclableMemoryStream.cs
```diff
@@ -514,26 +523,34 @@ public override Task CopyToAsync(Stream destination, int bufferSize, Cancellatio
...
-                            ++currentBlock;
-                        }
+                        return destination.WriteAsync(this.blocks[0], 0, this.length, cancellationToken);
                     }
                     else
                     {
-                        await destination.WriteAsync(this.largeBuffer, 0, this.length, cancellationToken);
+                        return CopyToAsyncImpl();
+
+                        async Task CopyToAsyncImpl()
```
benmwatson, line 536: I agree, this is a cleaner, better implementation.

benmwatson (changes requested):

UnitTests/Tests.cs
paulomorgado: OOPS!

src/RecyclableMemoryStream.cs
paulomorgado: Reverted!

You should add an [.editorconfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options) file to the solution to prevent these unvoluntary changes.

src/RecyclableMemoryStream.cs
paulomorgado: Reverted!

You should add an [.editorconfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options) file to the solution to prevent these unvoluntary changes.

UnitTests/Tests.cs
```diff
@@ -2518,6 +2544,24 @@ public void CopyToAsyncOneBlock()
...
+        public void CopyToAsyncOneBlockNonMemoryStream()
+        {
+            using (var stream = GetDefaultStream())
+            {
+                var buffer = GetRandomBuffer(DefaultBlockSize);
+                stream.Write(buffer, 0, buffer.Length);
+                var filename = Path.GetRandomFileName();
+                using (var fileStream = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite, DefaultBlockSize, FileOptions.Asynchronous))
+                {
+                    stream.CopyToAsync(fileStream).Wait();
```
Therzok, line 2557: Asynchronizing the method could avoid having to do a synchronuos wait here.

paulomorgado: I saw that, but didn't want to mess with it as part of this PR.

Does NUnit follow the same rule as XUnit and VSTest of Task returning methods for async test methods?

Therzok: Yup, it should just work by having the test method have a `Task`-like return type.

paulomorgado: Without knowing the threading model of NUnit, I can't make a call here.

If it's not running tests on multiple threads, this might be the most performant way.

benmwatson: I generally do not care about the async perf of unit tests, and don't like to get too advanced with NUnit features--keep it simple. I think the Wait is fine here. Correctness is more important.

benmwatson (approved):

brunom:
I agree to remove Task.Run, but currently CopyToAsyncImpl shouldn't be a local function to avoid an allocation on the happy path as described in https://devblogs.microsoft.com/premier-developer/the-performance-characteristics-of-async-methods/

paulomorgado:
@brunom, that blog post is too old (😄). It doesn't cover [static local functions](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#static-local-functions), which is what is being used here.

brunom:
I didn't know about static local functions. That would work, but CopyToAsyncImpl() isn't static at the moment: https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/cc68dd11a48a6491748767c6cf3796aceb03ccce/src/RecyclableMemoryStream.cs#L536

paulomorgado:
Oh! I recall it now. There is no closer because the local function only uses class members.

If it was static, everything would have to be passed as a parameter.

brunom:
It captures the cancellation token. We should pass it as a parameter, yes.

paulomorgado:
You're right! missed that one!

Sent PR to fix it: #116

PR #114, opened 2020-12-02
Remove 2 GB size limit on streams

### LINKED ISSUE #102: Plan to support more than 2GB?
Hello,

I'm from [Dicom Server](https://github.com/microsoft/dicom-server) , and our service is heavily using RecyclableMemoryStream for performance improvement. 

One issue we are facing is handling with >2GB file. Our service deals with medical images, and it happens some time the file size could be super big and exceed 2 GB.

I see the 2GB limitation from [here](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/5e6030d5bb9a71537d01e2dd421733206eb91020/src/RecyclableMemoryStream.cs#L63) , so wondering if there is a plan to support more than 2GB ?

Thanks
Peng Chen

benmwatson:
This PR removes the stream size limits throughout the library.

The main benefit of this is that you can now have streams of nearly any arbitrary size (theoretical maximum limit is 4GB * 4GB bytes).  

However, there are some major caveats:

~~1. This caused a change to a few public methods, which will trigger a major version bump.~~
2. .NET still has a maximum array length, documented [here](https://docs.microsoft.com/dotnet/api/system.array?view=netcore-3.1#remarks). Also note that .NET Framework and .NET Core may have different limits. 
3. Because of the maximum array length, once a contiguous buffer is required, the stream size will have this limit. Conversely, if a stream is already longer than this limit, trying to convert it to a contiguous buffer will fail.

The upshot of this is that once users consume this change, they may have to do some trivial code changes (`int` -> `long`), but if they are using "reasonable" streams, then there should be no other impact.

If you are using multi-GB streams, on the other hand, usage gets more complicated. Your code can no longer assume use of contiguous buffers.

I am not 100% convinced of this change. A few people have asked for it, but it adds complexity to usage. Hopefully, that complexity is pay-for-play.

An alternative is to change the current limit of 2 GB to 4 GB, but I'm not sure that resolves people's issues.

Resolves #102


UnitTests/Tests.cs
```diff
@@ -1027,7 +1013,7 @@ public void WriteByteUpdatesLength()
         public void SafeReadByteDoesNotUpdateStreamPosition()
         {
             var stream = this.GetRandomStream();
-            for (var i = 0; i < stream.Length; i++)
+            for (long i = 0; i < stream.Length; i++)
             {
```
paulomorgado, line 1017: Why the change? `var` assumes the correct type.

You can keep `var` if you use the correct literal: `0L`.

UnitTests/Tests.cs
```diff
@@ -1043,7 +1029,7 @@ public void SafeReadByteDoesNotDependOnStreamPosition()
             var buffer = this.GetRandomBuffer(stream.Capacity * 2);
             stream.Write(buffer, 0, buffer.Length);
 
-            for (var i = 0; i < stream.Length; i++)
+            for (long i = 0; i < stream.Length; i++)
             {
```
paulomorgado, line 1033: Why the change? `var` assumes the correct type.

You can keep `var` if you use the correct literal: `0L`.

UnitTests/Tests.cs
```diff
@@ -1065,7 +1051,7 @@ public void SafeReadByteCanBeUsedInParallel()
...
                 for (var i = 0; i < 1000; i++)
                 {
-                    var position = this.random.Next(0, bufferLength);
+                    long position = this.random.Next(0, bufferLength);
                     var byteRead = stream.SafeReadByte(ref position);
```
paulomorgado, line 1055: Why the change? `var` assumes the correct type.

UnitTests/Tests.cs
```diff
@@ -1226,13 +1212,13 @@ public void SafeReadCanBeUsedInParallel()
             {
                 for (var i = 0; i < 5; i++)
                 {
-                    var position = this.random.Next(0, bufferLength);
-                    var startPosition = position;
-                    var length = this.random.Next(0, bufferLength - position);
+                    long position = this.random.Next(0, bufferLength);
+                    long startPosition = position;
+                    var length = this.random.Next(0, (int)(bufferLength - position));
                     var readBuffer = new byte[length];
```
paulomorgado, line 1218: Why the change? `var` assumes the correct type.

You can keep `var` if you use the correct literal: `0L`.

benmwatson:
@paulomorgado Do you have any more comments on this?

csalada (approved):

src/RecyclableMemoryStream.cs
```diff
@@ -210,6 +231,18 @@ public RecyclableMemoryStream(RecyclableMemoryStreamManager memoryManager, Guid
         /// <param name="requestedSize">The initial requested size to prevent future allocations</param>
         /// <param name="initialLargeBuffer">An initial buffer to use. This buffer will be owned by the stream and returned to the memory manager upon Dispose.</param>
         internal RecyclableMemoryStream(RecyclableMemoryStreamManager memoryManager, Guid id, string tag, int requestedSize, byte[] initialLargeBuffer)
```
yabli, line 233: do we keep this for backcomp reason? even it's internal access?

paulomorgado: Agreed! It should be removed.

paulomorgado:
> @paulomorgado Do you have any more comments on this?

@benmwatson, nothing comes to mind at the moment.

src/RecyclableMemoryStream.cs
benmwatson: Will check on this. Might be a relic.

paulomorgado (approved):

paulomorgado (approved):

src/RecyclableMemoryStream.cs
```diff
@@ -610,7 +610,7 @@ async Task CopyToAsyncImpl()
                 }
                 else
                 {
-                    return destination.WriteAsync(this.largeBuffer, 0, this.length, cancellationToken);
+                    return destination.WriteAsync(this.largeBuffer, 0, (int)this.length, cancellationToken);
```
paulomorgado, line 613: We should add `Debug.Assert` assertions everywhere `length` is assumed to be in the `int` boundary, just in case.

paulomorgado (approved):

paulomorgado (approved):

paulomorgado (approved):

paulomorgado (approved):

Issue #115, opened 2020-12-07
CopyToAsync does not take current position into account


DavidZidar:
The regular `MemoryStream` copies data from the current position and so does RecyclableMemoryStream version 1.3.5 but version 1.3.6 copies from the beginning regardless of position which is unexpected.

This breaks my use case because I often read stream headers and then copy the remainder.

The problem can easily be reproduced like this:
```
var manager = new RecyclableMemoryStreamManager();
using var input = manager.GetStream();
input.WriteByte(1);
input.WriteByte(2);
input.Seek(1, SeekOrigin.Begin);
using var output = new MemoryStream();
await input.CopyToAsync(output);

output.Length // Should now be 1, but is 2 with version 1.3.6
```

benmwatson:
Thanks for reporting it. We've been changing how CopyToAsync is implemented. I'll look into this soon.

benmwatson:
@DavidZidar and @paulomorgado Could you please take a look at the related PR to resolve this?

PR #118, opened 2021-01-08
Fix CopyToAsync to start from current position

benmwatson:
Issue #115 points out that `CopyToAsync` should start from the current stream position, not 0. Looking at the `MemoryStream` code, I confirm this.

This PR corrects this oversight in all code paths, updates the unit tests to catch this, and regenerates documentation to make this clear.

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -2490,17 +2490,19 @@ namespace Microsoft.IO.UnitTests
             }            
         }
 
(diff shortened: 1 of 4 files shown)


DavidZidar (approved):
Looks good to me. Test coverage seems good and I can confirm it passes my test suite as well, the previous version did not.

paulomorgado (approved):

PR #122, opened 2021-01-19
Drop support for netstandard1.4 and net40

### LINKED ISSUE #120: Removing net40 and netstandard1.4
I am preparing to release v2.0 of RecyclableMemoryStream. In addition to removing the 2 GB stream limit, one of the main changes will be removing support for net40 and netstandard1.4. 

The main reason is that these two targets have significant limitations which can cause nonobvious behavior in some circumstances.

If this would significantly impact you, please let me know.

benmwatson:
These platforms are not needed, and are largely obsolete.

* netstandard1.4 is problematic because it doesn't support some critical methods correctly (e.g., `Close` and `GetBuffer`).
* net40 doesn't support ETW and other newer runtime features
~~* net46 doesn't appear to have ever been needed--there was no net46-specific code present.~~ Keeping net46 for now because there is net45-specific code that is less efficient. No reason to punish users of newer frameworks.

Resolves #120


README.md
```diff
@@ -36,7 +36,9 @@ Microsoft.IO.RecyclableMemoryStream is a MemoryStream replacement that offers su
 
 At least MSBuild 15 is required to build the code. You get this with Visual Studio 2017.
 
-Build targets are: net40, net45, netstandard1.4, netstandard2.1, and netcoreapp2.1.
+Build targets in v2 are: net45, netstandard2.1, and netcoreapp2.1 (net40 and netstandard1.4 were deprecated).
```
erikness, line 39: v2 - planning a major version bump for this?

erikness (commented):
According to the [.NET standard chart](https://docs.microsoft.com/en-us/dotnet/standard/net-standard), if you were *just* targeting multiple versions of .NET Standard, you'd be leaving out .NET Framework 4.6.0 users and below with this change.

(It says .NET Framework 4.6.1 corresponds to .NET Standard 1.4, but with some footnotes that it can also use up to .NET Standard 2.0.)

Keeping support for .NET Framework 4.5 seems to be enough if some teams are still using those old versions of .NET Framework. So this seems safe to me. And people can always use old versions or create an issue if they really need support for 5+ years old versions of .NET.

README.md
erikness: ah yes, I just saw it in the other commit.

yabli, line 39: netstandard2.0 ?

yabli (approved):

paulomorgado (approved):

benmwatson:
@erikness I've added back net46 because there was less-efficient net45-specific code. No reason to punish net46 and up.

PR #128, opened 2021-01-25
Remove use of Interlocked methods from Dispose and Disposed.

### LINKED ISSUE #111: Why is disposed state using Interlocked?
This is being used in all invocations of the stream.

is there any concurrent work going on that I'm not aware of?

benmwatson:
Resolve #111.

The `Interlocked` calls weren't really useful--streams should never be used in a concurrent manner anyway, and their usage negatively impacts the `CheckDisposed` calls.

I've also updated the documentation to make it clear that concurrent stream access is not supported.


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+# Version 2.0
+
+* Removed 2 GB stream limit. Extremely large streams can be created by chaining together blocks. Attempts to convert a stream into a contiguous buffer longer than the runtime's limits will result in an exception.
+* Removed build targets for net40, net462, and netstandard1.4.
```
erikness, line 4: 1. Technically, you removed `net46` instead of `net462`. I'm not that familiar with how much the build system cares about that third part of the .NET Framework version - is this an important distinction to make?
1. Maybe mention that we still build for net45 - this sentence alone would kind of make me suspect that there's no longer support for applications on .NET 4.6.2.
    1. (though even if we didn't have a net45 build target, applications using .NET Framework 4.6.2 would still be able to depend on the netstandard2.0 build. But that's probably not worth mentioning in the change log, people can figure that out themselves).

erikness (commented):
Are there any circumstances you can think of where Dispose(true) and Dispose(false) would run concurrently? It comes to mind because .NET finalization is not controlled by application developers so it might be trickier to avoid concurrency with that.

I can't come up with a scenario. To call Dispose(true), you need a reference until that call is finished, and if you have a reference, then the GC won't collect and there won't be finalization. Even if you have a WeakReference to your MemoryStream, you've gotta make a new strong reference before calling any methods on it.

yabli:
looks good to me

paulomorgado (approved):

PR #129, opened 2021-01-28
Refactor .NET events to provide same info as ETW events

### LINKED ISSUE #125: Adding the stream guid and tag to the Stream Finalized C# event
Currently the C# events for stream finalization, disposal and so on, do not receive the string id and tag of the streams finalized as arguments. They also do not receive the size. This prevents someone from writing code dependent on the stream which is being finalized.

Some potential uses that this would enable are:
1) Hooking up a custom telemetry system to monitor leaks by using the tags or the guids and emitting from C#
2) Using the string tag as a category and implementing per application area statistics
3) Unique events per stream type

A suggestion would be to add these arguments as they're available at the point that these events are emitted and they would enable a lot of uses for the events.

benmwatson:
When the .NET events were first implemented, they were done mainly to provide simple hooks for reporting performance counters in Bing. With the expanded user base, it makes sense to bring these to a usefulness level on-par with the ETW events for those who can't leverage ETW for whatever reason.

Design constraints:
* Pay-for-play. If you don't subscribe to these events, they cost nothing (well, a null delegate check, same as before).
* Standardized argument order, refactored ETW events to match.
* Added more info in most events.
* Refactored code to put ETW and .NET events in the same place.

Note that there is not a 1:1 correspondence of ETW and .NET events. A few matches didn't seem to make sense. e.g., `ReportUsage`--this is a very common event and is useful mostly for performance counters, but would quickly spam a log file.

Resolves #125


src/RecyclableMemoryStream.cs
```diff
@@ -540,7 +532,7 @@ public override byte[] GetBuffer()
 
             if (this.blocks.Count > 0 && this.memoryManager.AggressiveBufferReturn)
             {
-                this.memoryManager.ReturnBlocks(this.blocks, this.tag);
+                this.memoryManager.ReturnBlocks(this.blocks, Guid.Empty, this.tag);
```
erikness, line 535: Guid.Empty? This RMS still has an id, why can't we tell that to the RMS manager when we return these blocks?

src/RecyclableMemoryStreamManager.cs
```diff
@@ -538,53 +502,74 @@ internal void ReturnBlocks(ICollection<byte[]> blocks, string tag)
...
+            this.BlockCreated?.Invoke(new BlockCreatedEventArgs(this.smallPoolInUseSize));
         }
 
-        internal void ReportBlockDiscarded()
+        internal void ReportLargeBufferCreated(long requiredSize, bool pooled, Guid id, string tag, string callStack)
         {
-            this.BlockDiscarded?.Invoke();
+            if (pooled)
+            {
+                Events.Writer.MemoryStreamNewLargeBufferCreated(requiredSize, this.LargePoolInUseSize);
```
erikness, line 513: id + tag?

erikness (changes requested):
One more thing: in the README, update the list of event hooks:

1. LargeBufferCreatd => LargeBufferCreated (typo)
2. LargeBufferDiscarded => BufferDiscarded

src/RecyclableMemoryStream.cs
benmwatson: Woops, good catch

src/RecyclableMemoryStreamManager.cs
benmwatson: they are both there already, but I didn't normalize their order in the param list like I did elsewhere. Fixing.

src/EventArgs.cs
```diff
@@ -0,0 +1,436 @@
+﻿namespace Microsoft.IO
+{
+    using System;
+
+    public sealed partial class RecyclableMemoryStreamManager
+    {
+        /// <summary>
+        /// Arguments for the StreamCreated event
+        /// </summary>
+        public struct StreamCreatedEventArgs
```
Therzok, line 10: Would making these structs readonly be a limiting factor for future API updates? Otherwise, I think these are all candidates to be readonly.

src/RecyclableMemoryStreamManager.cs
```diff
@@ -372,24 +342,22 @@ internal byte[] GetLargeBuffer(long requiredSize, string tag)
...
                 buffer = new byte[requiredSize];
-                string callStack = null;
                 if (this.GenerateCallStacks)
                 {
                     // Grab the stack -- we want to know who requires such large buffers
                     callStack = Environment.StackTrace;
-                }
-                Events.Writer.MemoryStreamNonPooledLargeBufferCreated(requiredSize, tag, callStack);
-                ReportLargeBufferCreated();
+                }
```
paulomorgado, line 356: This is a tricky one, but, even if `GenerateCallStacks` is true, this might be a pooled buffer or the events aren't being subscribed.

It would be better to instantiate a `StackTrace` instance with the necessary `skipFrames`.

Doesn't `EventSource` provide stack trace already?

src/RecyclableMemoryStreamManager.cs
```diff
@@ -801,51 +786,56 @@ public MemoryStream GetStream(string tag, Memory<byte> buffer)
...
+        /// Triggered when a stream is requested to expand beyond the maximum length specified by the responsible RecyclableMemoryStreamManager.
+        /// </summary>
+        public event Action<StreamOverCapacityEventArgs> StreamOverCapacity;
 
         /// <summary>
-        /// Triggered when a large buffer is discarded, along with the reason for the discard.
+        /// Triggered when a buffer of either type is discarded, along with the reason for the discard.
         /// </summary>
-        public event LargeBufferDiscardedEventHandler LargeBufferDiscarded;
+        public Action<BufferDiscardedEventArgs> BufferDiscarded;
```
paulomorgado, line 834: Why isn't this an event anymore?

src/RecyclableMemoryStream.cs
```diff
@@ -699,7 +691,7 @@ public override byte[] ToArray()
             this.CheckDisposed();
 
             string stack = this.memoryManager.GenerateCallStacks ? Environment.StackTrace : null;
```
paulomorgado, line 693: Stack trace should be generated only if really needed.

src/RecyclableMemoryStreamManager.cs
benmwatson: woops, oversight

src/RecyclableMemoryStream.cs
benmwatson: That's what it is doing? Are you saying that it should only record stack trace if ThrowExceptionOnToArray is also true? There is a valid scenario where I still want to allow it to go through, but want it logged to debug and remove.

src/RecyclableMemoryStreamManager.cs
paulomorgado: Any reason these aren't `EventHandler<TEventArgs>`?

src/RecyclableMemoryStream.cs
paulomorgado: Unless I checked all code paths wrong, there's the chance a stack trace will be generated for nothing.

PR #137, opened 2021-02-19
Unseal RecyclableMemoryStream to enable extending/overriding methods

### LINKED ISSUE #136: Cannot derive from sealed type 'RecyclableMemoryStream'
This library advertises itself as a drop-in replacement for Stream, but is marked as `sealed`, which means I cannot create a
```cs
internal class ReadOnlyStream: Microsoft.IO.RecyclableMemoryStream
```

Since there is no `IStream` interface to be implemented, we'll need a more meaningful refactor to migrate from `Stream` to `RecyclableMemoryStream`.

NateRadebaugh:
I found this succinct explanation for why this class may be marked as sealed:
> https://stackoverflow.com/a/52907478/354130
> 
> Marking a class as `Sealed` prevents tampering of important classes that can compromise security, or affect performance.
>
> Many times, sealing a class also makes sense when one is designing a utility class with fixed behaviour, which we don't want to change.

However, I'd like to extend this class to create a `ReadOnlyStream` and utilize the same properties as a normal `Stream`

---

Fixes #136

### DIFF
--- a/docs/Microsoft.IO/RecyclableMemoryStream.md
+++ b/docs/Microsoft.IO/RecyclableMemoryStream.md
@@ -3,7 +3,7 @@
 MemoryStream implementation that deals with pooling and managing memory streams which use potentially large buffers.
 
 ```csharp
-public sealed class RecyclableMemoryStream : MemoryStream
+public class RecyclableMemoryStream : MemoryStream
 ```
 
 ## Public Members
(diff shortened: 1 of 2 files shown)


paulomorgado:
Making a class sealed or not is a design choice that has implications in terms of reliability and performance.

You can always use [delegation](https://en.wikipedia.org/wiki/Delegation_pattern) instead of [inheritance](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)).

PR #138, opened 2021-02-22
Implement IBufferWriter<byte>

grbell-ms:
This PR allows users of RMS to write directly to the underlying memory without the limitations of `GetBuffer()` (copying, and failing on streams larger than 2GB).

The only real complexity is handling calls to `GetMemory()` and `GetSpan()` with a non-zero size hint. If the hint is larger than the space left in the current block the methods rent and return a temporary buffer from the memory manager. A later call to `Advance()` copies the data into the stream and returns the buffer.


src/RecyclableMemoryStream.cs
```diff
@@ -619,11 +623,126 @@ async Task CopyToAsyncImpl(CancellationToken ct)
...
+        /// </summary>
+        private ArraySegment<byte> GetWritableBuffer(int sizeHint)
+        {
+            this.CheckDisposed();
+            if (sizeHint < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(sizeHint), $"{nameof(sizeHint)} must be non-negative");
+            }
+
+            sizeHint = Math.Max(sizeHint, 1);
```
paulomorgado, line 714: Is `1` an sensible default?

Consider providing overloads without sizeHint specification that will return the rest of the current buffer.

paulomorgado, line 714: Given that, at this point, `sizeHint` can't be less than `0`, this line of code is functionally equivalent to:

```CSharp
if (sizeHint == 0)
{
    sizeHint = 1;
}
```

With a better performance footprint.

src/RecyclableMemoryStream.cs
```diff
@@ -1310,13 +1429,21 @@ private BlockAndOffset GetBlockAndRelativeOffset(long offset)
...
 
-        private void EnsureCapacity(long newCapacity)
+        private void EnsureCapacity(long newCapacity, bool throwOomExceptionOnOverCapacity = false)
         {
             if (newCapacity > this.memoryManager.MaximumStreamCapacity && this.memoryManager.MaximumStreamCapacity > 0)
             {
                 this.memoryManager.ReportStreamOverCapacity(this.id, this.tag, newCapacity, this.AllocationStack);
-                throw new InvalidOperationException("Requested capacity is too large: " + newCapacity + ". Limit is " +
-                                                    this.memoryManager.MaximumStreamCapacity);
+                string message = $"Requested capacity is too large: {newCapacity}. Limit is {this.memoryManager.MaximumStreamCapacity}";
```
paulomorgado, line 1437: Nitpicking: this has 2 boxing operations.

For 4 blocks, this is the best option:

```CSharp
string message = "Requested capacity is too large: " + newCapacity.ToString(CultureInfo.InvariantCulture) + ". Limit is " + this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture);
```

But, in the general case, prefer:

```CSharp
string message = $"Requested capacity is too large: {newCapacity.ToString(CultureInfo.InvariantCulture)}. Limit is {this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture)}";
```

src/RecyclableMemoryStream.cs
grbell-ms: This method does return the rest of the current buffer by default. `sizeHint` is the minimum size of the returned buffer, which is at least 1.

I don't think we need to expand the API surface beyond `IBufferWriter`

grbell-ms: I've renamed `sizeHint` to clarify the issue.

src/RecyclableMemoryStream.cs
grbell-ms: I don't expect this code path to be hit very much, but it's an easy fix.

src/RecyclableMemoryStream.cs
```diff
@@ -1310,13 +1429,21 @@ private BlockAndOffset GetBlockAndRelativeOffset(long offset)
...
+
+                if (throwOomExceptionOnOverCapacity)
```
benmwatson, line 1445: hmm...not sure how I feel about this. I see the expectations of the interface, and they conflict with the existing implementation. I might consider just making it OOM here officially. This is going to go into v2.0, which can have some breaking changes, and this would be one.

benmwatson (commented):
Please also update the readme.md document with instructions on how to use this, maybe a link to IBufferWriter documentation.

src/RecyclableMemoryStream.cs
```diff
@@ -619,11 +623,126 @@ async Task CopyToAsyncImpl(CancellationToken ct)
...
+        /// Seeks forward by <paramref name="count"/> bytes.
+        /// </summary>
+        /// <remarks>
+        /// You must request a new buffer after calling Advance to continue writing more data and cannot write to a previously acquired buffer.
+        /// </remarks>
+        /// <param name="count">How many bytes to advance</param>
+        /// <exception cref="ObjectDisposedException">Object has been disposed</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="count"/> is negative</exception>
+        /// <exception cref="InvalidOperationException"><paramref name="count"/> is larger than the size of the previously requested buffer</exception>
+        public void Advance(int count)
```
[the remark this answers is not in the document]
paulomorgado: I thought about that. But then I also thought it wouldn't be polite to force users of both APIs to have to cast to use one or the other.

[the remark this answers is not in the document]
grbell-ms: I think that depends on what this API is for. If we want to replace `GetBuffer()` calls we should keep the implicit implementation. If we just want to interoperate with APIs that take `IBufferWriter<byte>` we should go with an explicit implementation.

I think implicit is the right move here.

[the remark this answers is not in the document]
paulomorgado: Explicit implementations are for hiding the implementation from the use of a class or when more than one implemented interface defines the same member.

For example, `List<T>` implements `IEnumerable<T>` that implements `IEnumerable`. `IEnumerable<T>` is implemented implicitly and `IEnumerable` is implemented explicitly.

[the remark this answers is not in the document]
grbell-ms: Right, we want the `IBufferWriter<byte>` methods to be implicitly implemented because we want consumers of RMS to use them directly.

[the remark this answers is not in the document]
grbell-ms: Thinking about it now, do we want to add a `ThrowExceptionOnGetBuffer` property to the stream manager?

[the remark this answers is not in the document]
paulomorgado: Large buffers are needed for backward compatibility.

But, with this implementation, one of the use cases for large buffers (calling `ToArray` and `GetBuffer` is no longer needed.

grbell-ms:
@benmwatson How do the instructions look?

benmwatson (approved):

benmwatson:
> @benmwatson How do the instructions look?

very nice. Thank you! I've approved this and will merge it in soon. I have some more documentation fixes I want to do next week, and I'll plan on releasing 2.0 late next week.

PR #140, opened 2021-03-10
Fix bug with CopyToAsync not adjusting source position

### LINKED ISSUE #139: CopyToAsync reset source stream position
We found strange behaviour change between 1.4.0 and previous version at **CopyToAsync** method. At version 1.4.0 this method reset (IMO unexpectedly) the SOURCE stream position back to position before call.

benmwatson:
Resolves #139 

The RMS version incorrectly kept the stream's position at the pre-copy value, when the correct behavior according to `MemoryStream`s implementation is to in fact increment the position as if it were a read operation.

This PR resolves this bug and adds some unit tests to ensure all code paths are hit.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -566,6 +566,7 @@ namespace Microsoft.IO
         ///   <paramref name="destination" /> is <see langword="null" />.</exception>
         /// <exception cref="T:System.ObjectDisposedException">Either the current stream or the destination stream is disposed.</exception>
         /// <exception cref="T:System.NotSupportedException">The current stream does not support reading, or the destination stream does not support writing.</exception>
+        /// <remarks>Similarly to <c>MemoryStream</c>'s behavior, <c>CopyToAsync</c> will adjust the source stream's position by the number of bytes written to the destination stream, as a Read would do.</remarks>
         public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
         {
             if (destination == null)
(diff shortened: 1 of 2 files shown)


src/RecyclableMemoryStream.cs
```diff
@@ -592,15 +596,16 @@ public override Task CopyToAsync(Stream destination, int bufferSize, Cancellatio
                     if (this.blocks.Count == 1)
                     {
                         AssertLengthIsSmall();
-                        return destination.WriteAsync(this.blocks[0], (int)this.position, (int)(this.length - this.position), cancellationToken);
+                        return destination.WriteAsync(this.blocks[0], (int)this.position, (int)count, cancellationToken).ContinueWith((task) => this.position += count);
```
grbell-ms, line 599: We can avoid creating a continuation by advancing the position first. (This is what MemoryStream does.)
```suggestion
                        var pos = this.position;
                        this.position += count;
                        return destination.WriteAsync(this.blocks[0], (int)pos, (int)count, cancellationToken);
```

src/RecyclableMemoryStream.cs
```diff
@@ -612,13 +617,14 @@ async Task CopyToAsyncImpl(CancellationToken ct)
...
                             }
+                            this.position += totalBytesToWrite;
                         }
                     }
                 }
                 else
                 {
                     AssertLengthIsSmall();
-                    return destination.WriteAsync(this.largeBuffer, (int)this.position, (int)(this.length - this.position), cancellationToken);
+                    return destination.WriteAsync(this.largeBuffer, (int)this.position, (int)count, cancellationToken).ContinueWith((task) => this.position += count);
```
grbell-ms, line 627: Same thing here.

PR #145, opened 2021-03-11
Add WriteTo(byte[]....) method overloads

benmwatson:
These methods just add yet more flexibility in getting data out of the stream without resorting to `GetBuffer` calls.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -972,7 +972,7 @@ namespace Microsoft.IO
         }
         
 #endif
-
+        
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/RecyclableMemoryStream.cs
```diff
@@ -972,7 +972,7 @@ public int SafeRead(Span<byte> buffer, ref long streamPosition)
         }
         
 #endif
-
+
```
paulomorgado, line 975: Unnecessary white space.

paulomorgado (approved):

paulomorgado (approved):

PR #146, opened 2021-03-11
Update README, CHANGES, and API docs

benmwatson:
Adding latest changes, editing the readme significantly, re-generating API documentation.


README.md
```diff
@@ -52,27 +54,25 @@ Read the change log [here](CHANGES.md).
...
 
-The MemoryStreamManager class maintains two separate pools of objects:
+The `RecyclableMemoryStreamManager` class maintains two separate pools of objects:
 
-1. **Small Pool** - Holds small buffers (of configurable size). Used by default for all normal read/write operations. Multiple small buffers are chained together in the RecyclableMemoryStream class and abstracted into a single stream.
-2. **Large Pool** - Holds large buffers, which are only used when you must have a single, contiguous buffer, such as when you plan to call GetBuffer(). 
+1. **Small Pool** - Holds small buffers (of configurable size). Used by default for all normal read/write operations. Multiple small buffers are chained together in the `RecyclableMemoryStream` class and abstracted into a single stream.
+2. **Large Pool** - Holds large buffers, which are only used when you must have a single, contiguous buffer, such as when you plan to call `GetBuffer()`. It is possible to create streams longer than is possible to be represented by a single buffer because of .NET's array size limits.
```
paulomorgado, line 62: **longer** or **larger**?

README.md
```diff
@@ -110,7 +110,7 @@ using (var stream = manager.GetStream("Program.Main"))
 }
 ```
 
-You can also provide an existing buffer. It’s important to note that this buffer will be copied into the pooled buffers owned by RecyclableMemoryStream:
+You can also provide an existing buffer. It’s important to note that this buffer will be *copied* into the pooled buffers owned by `RecyclableMemoryStream`:
```
paulomorgado, line 113: Does this mean that the buffer will be added to the pool or that the contents of the buffer will be copied to a buffer from the pool?

README.md
```diff
@@ -134,11 +134,13 @@ manager.MaximumFreeLargePoolBytes = maxBufferSize * 4;
...
 ```
 
+### Usage Guidelines
+
 While this library strives to be very general and not impose too many restraints on how you use it, its purpose is to reduce the cost of garbage collections incurred by frequent large allocations. Thus, there are some general guidelines for usage that may be useful to you:
 
 1. Set the `blockSize`, `largeBufferMultiple`, `maxBufferSize`, `MaximumFreeLargePoolBytes` and `MaximumFreeSmallPoolBytes` properties to reasonable values for your application and resource requirements.
 2. Always dispose of each stream exactly once.
-3. Never call `ToArray` and avoid calling `GetBuffer` if possible. Instead, use `GetReadOnlySequence` for reading and `GetSpan`\\`GetMemory` with `Advance` for writing.
+3. Never call `ToArray` and avoid calling `GetBuffer` if possible. Instead, use `GetReadOnlySequence` for reading and `GetSpan`\\`GetMemory` with `Advance` for writing. There are also miscellaneous `CopyTo` and `WriteTo` methods that may be convenient.
```
paulomorgado, line 143: I think it would be nice to mention that `GetSpan`\\`GetMemory` and `Advance` are from the implementation of `IBufferWriter<byte>`-

Maybe:

> `IBufferWriter<byte>.GetSpan`\\`IBufferWriter<byte>.GetMemory` with `IBufferWriter<byte>.Advance`

paulomorgado (approved):

PR #163, opened 2021-06-03
Fix typo in code sample

### LINKED ISSUE #154: Documentation Suggestion - GetReadOnlySequence readme.md example incorrect
The example code for GetReadOnlySequence in the root readme.md appears to be incorrect:

```csharp
using (var stream = manager.GetStream())
using (var sha256Hasher = IncrementalHash.CreateHash(HashAlgorithmName.SHA256))
{
    foreach (var memory in stream.GetReadOnlySequence()) // Stream is a MemoryStream NOT a RecyclableMemoryStream
    {
        sha256Hasher.AppendData(memory.Span);
    }
....
```
The stream is an MemoryStream and so we'd need to cast as a RecyleableMemoryStream for the call to `GetReadonlySequence()`.  Or am I missing something?

benmwatson:
Resolves #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ When configuring the options, consider questions such as these:
 ```
 var bigInt = BigInteger.Parse("123456789013374299100987654321");
 
-using (var stream = manager.GetStream())
+using (var stream = manager.GetStream() as RecyclableMemoryStream)
 {
     Span<byte> buffer = stream.GetSpan(bigInt.GetByteCount());
     bigInt.TryWriteBytes(buffer, out int bytesWritten);
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ using (var stream = manager.GetStream())
 `GetReadOnlySequence` returns a [ReadOnlySequence<byte>](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.readonlysequence-1?view=netstandard-2.1) that can be used for zero-copy stream processing. For example, hashing the contents of a stream: 
 
 ```
-using (var stream = manager.GetStream())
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,7 +173,7 @@ using (var stream = manager.GetStream())
 `GetReadOnlySequence` returns a [ReadOnlySequence<byte>](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.readonlysequence-1?view=netstandard-2.1) that can be used for zero-copy stream processing. For example, hashing the contents of a stream: 
 
 ```
-using (var stream = manager.GetStream())
+using (var stream = manager.GetStream() as RecyclableMemoryStream)
```
grbell-ms, line 176: The code on line 163 has the same issue.

grbell-ms (approved):

README.md
benmwatson: Thanks, good catch.

paulomorgado (approved):
Maybe we should change `GetStream` to return `RecyclableMemoryStream` instead.

There's no point in returing `MemoryStream`.

grbell-ms:
> Maybe we should change `GetStream` to return `RecyclableMemoryStream` instead.
> 
> There's no point in returing `MemoryStream`.

I believe the change would maintain binary compatibility.

benmwatson:
> > Maybe we should change `GetStream` to return `RecyclableMemoryStream` instead.
> > There's no point in returing `MemoryStream`.
> 
> I believe the change would maintain binary compatibility, so we should definitely return RecyclableMemoryStream.

Definitely worth considering. I'll add an issue to track this.

PR #173, opened 2021-06-15
Fix integer overflow bug in RoundToLargeBufferSize

### LINKED ISSUE #171: Infinite loop in RoundToLargeBufferSize if requiredSize > 1gb and UseExponentialLargeBuffer == true
Introduced here: https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/commit/d9d99a4f8314693f89c32e6e4e8efab5839f40d9#diff-731ca724bbb85a55c0a5a4998af68010f944bafaf55eea25877ab26372b580a3R402

The while loop 

```
int pow = 1;
while (this.LargeBufferMultiple * pow < requiredSize)
{
    pow <<= 1;
}
```

compares the value of `LargeBlockMultiple * pow` to `requiredSize`. When `requiredSize` was widened, `LargeBlockMultiple` and `pow` were left as `int` which allows an overflow to occur. 

Test `AllMultiplesOrExponentialUpToMaxCanBePooled` will cover this if MaxBufferSize is initialized to something like:
`const int MaxBufferSize = (int.MaxValue / LargeBufferMultiple) * LargeBufferMultiple;`

benmwatson:
Resolve #171 

This fixes an integer overflow problem that occurs now that sizes can be `long` values. It adds a test that verifies the issue is fixed and upgrades the NUnit packages in order to use the `Timeout` attribute.

### DIFF
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -457,7 +457,7 @@ namespace Microsoft.IO
         {
             if (this.UseExponentialLargeBuffer)
             {
-                int pow = 1;
+                long pow = 1;
                 while (this.LargeBufferMultiple * pow < requiredSize)
                 {
                     pow <<= 1;
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -3596,6 +3596,19 @@ namespace Microsoft.IO.UnitTests
             }
         }
 
+        // Issue #171 - Infinite Loop when particular values are chosen causing an int overflow during
+        // validation of buffer sizes.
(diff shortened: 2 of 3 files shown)


src/RecyclableMemoryStreamManager.cs
```diff
@@ -457,7 +457,7 @@ private long RoundToLargeBufferSize(long requiredSize)
         {
             if (this.UseExponentialLargeBuffer)
             {
-                int pow = 1;
+                long pow = 1;
```
cryolithic, line 460: This caught me up for a bit as I still write a lot of Windows c++. Hopefully I can submit as PR next time, still waiting on legal to clear open source contribs for my team 🙄

cryolithic (approved):

PR #192, opened 2021-07-28
Improve documentation

hankovich:
There are a couple of issues in the documentation (both README and xml-comments). Some of the docs are outdated and not valid anymore, some of them are not consistent.

I tried to find all issues and inconsistencies and address them.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -1535,10 +1535,8 @@ namespace Microsoft.IO
             if (newCapacity > this.memoryManager.MaximumStreamCapacity && this.memoryManager.MaximumStreamCapacity > 0)
             {
                 this.memoryManager.ReportStreamOverCapacity(this.id, this.tag, newCapacity, this.AllocationStack);
-
-                var requestedCapacity = newCapacity.ToString(CultureInfo.InvariantCulture);
-                var limit = this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture);
-                throw new OutOfMemoryException($"Requested capacity is too large: {requestedCapacity}. Limit is {limit}.");
+                
+                throw new OutOfMemoryException($"Requested capacity is too large: {newCapacity}. Limit is {this.memoryManager.MaximumStreamCapacity}.");
             }
 
             if (this.largeBuffer != null)
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -1427,7 +1427,7 @@ namespace Microsoft.IO
         #region Helper Methods
         private bool Disposed => this.disposed;
 
-        [MethodImpl((MethodImplOptions)256)]
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void CheckDisposed()
         {
             if (this.Disposed)
@@ -1524,7 +1524,7 @@ namespace Microsoft.IO
             }
         }
 
-        [MethodImpl((MethodImplOptions)256)]
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private BlockAndOffset GetBlockAndRelativeOffset(long offset)
         {
             var blockSize = this.memoryManager.BlockSize;
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -683,31 +683,28 @@ namespace Microsoft.IO
         /// IMPORTANT: Calling Write(), GetBuffer(), TryGetBuffer(), Seek(), GetLength(), Advance(),
         /// or setting Position after calling GetMemory() invalidates the memory.
         /// </remarks>
-        public Memory<byte> GetMemory(int sizeHint = 0) => this.GetWritableBuffer(sizeHint, nameof(sizeHint));
+        public Memory<byte> GetMemory(int sizeHint = 0) => this.GetWritableBuffer(sizeHint);
 
         /// <inheritdoc/>
         /// <remarks>
         /// IMPORTANT: Calling Write(), GetBuffer(), TryGetBuffer(), Seek(), GetLength(), Advance(),
         /// or setting Position after calling GetSpan() invalidates the span.
         /// </remarks>
-        public Span<byte> GetSpan(int sizeHint = 0) => this.GetWritableBuffer(sizeHint, nameof(sizeHint));
+        public Span<byte> GetSpan(int sizeHint = 0) => this.GetWritableBuffer(sizeHint);
 
         /// <summary>
         /// When callers to GetSpan() or GetMemory() request a buffer that is larger than the remaining size of the current block
         /// this method return a temp buffer. When Advance() is called, that temp buffer is then copied into the stream.
         /// </summary>
-        private ArraySegment<byte> GetWritableBuffer(int minimumBufferSize, string originalParamName)
+        private ArraySegment<byte> GetWritableBuffer(int sizeHint)
         {
             this.CheckDisposed();
-            if (minimumBufferSize < 0)
+            if (sizeHint < 0)
             {
-                throw new ArgumentOutOfRangeException(originalParamName, $"{originalParamName} must be non-negative.");
+                throw new ArgumentOutOfRangeException(nameof(sizeHint), $"{nameof(sizeHint)} must be non-negative.");
             }
 
-            if (minimumBufferSize == 0)
-            {
-                minimumBufferSize = 1;
-            }
+            var minimumBufferSize = Math.Max(sizeHint, 1);
 
             this.EnsureCapacity(this.position + minimumBufferSize);
             if (this.bufferWriterTempBuffer != null)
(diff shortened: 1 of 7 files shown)


src/RecyclableMemoryStream.cs
```diff
@@ -706,7 +701,7 @@ private ArraySegment<byte> GetWritableBuffer(int minimumBufferSize)
             this.CheckDisposed();
             if (minimumBufferSize < 0)
             {
-                throw new ArgumentOutOfRangeException("sizeHint", $"sizeHint must be non-negative");
+                throw new ArgumentOutOfRangeException(nameof(minimumBufferSize), $"{nameof(minimumBufferSize)} must be non-negative.");
```
grbell-ms, line 704: The parameter name exposed to callers is `sizeHint`, not `minimumBufferSize`.

src/RecyclableMemoryStream.cs
```diff
@@ -1530,7 +1540,7 @@ private void EnsureCapacity(long newCapacity)
                 this.memoryManager.ReportStreamOverCapacity(this.id, this.tag, newCapacity, this.AllocationStack);
                 throw new OutOfMemoryException(
                     "Requested capacity is too large: " + newCapacity.ToString(CultureInfo.InvariantCulture) +
-                    ". Limit is " + this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture));
+                    ". Limit is " + this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture) + ".");
```
grbell-ms, line 1543: Let's make this one an interpolated string too.

grbell-ms: And calling the `ToString()` override is also unnecessary.

src/RecyclableMemoryStream.cs
```diff
@@ -403,7 +399,7 @@ public override int Capacity
                 long size = (long)this.blocks.Count * this.memoryManager.BlockSize;
                 if (size > int.MaxValue)
                 {
-                    throw new InvalidOperationException("Capacity is larger than int.MaxValue. Use Capacity64 instead");
+                    throw new InvalidOperationException("Capacity is larger than int.MaxValue. Use Capacity64 instead.");
```
grbell-ms, line 402: Want to add some `nameof()` here?

src/RecyclableMemoryStream.cs
```diff
@@ -999,17 +993,17 @@ public override void Write(byte[] buffer, int offset, int count)
             if (offset < 0)
             {
                 throw new ArgumentOutOfRangeException(nameof(offset), offset,
-                                                      "Offset must be in the range of 0 - buffer.Length-1");
+                                                      $"{nameof(offset)} must be in the range of 0 - {nameof(buffer)}.Length-1.");
```
grbell-ms, line 996: Should we give `.Length` the `nameof()` treatment as well?

grbell-ms (commented):
This looks good! Just a few nit comments. It would have been easier to review if the whitespace/period changes and more substantial modifications were in separate commits, but I do appreciate you fixing all the oversights and formatting issues.

src/RecyclableMemoryStream.cs
paulomorgado: That will guarantee that docs won't be out of sync with whatever is being used at the moment.

src/RecyclableMemoryStream.cs
paulomorgado: `ToString` will avoid boxing of the value type.

And to use `CultureInfo.InvariantCulture`, you'd need to use `FormattableString.Invariant($"...")`.

hankovich:
> It would have been easier to review if the whitespace/period changes and more substantial modifications were in separate commits, but I do appreciate you fixing all the oversights and formatting issues.

I didn't know how deep this rabbit hole with whitespaces/periods was :)

src/RecyclableMemoryStream.cs
hankovich: Ah, you're right. I've refactored this place a bit.

src/RecyclableMemoryStream.cs
hankovich: @paulomorgado  Didn't know about the `FormattableString.Invariant($"...")`, thanks!
But does it seem overcomplicated to think about allocating memory in a location that throws the OutOfMemoryException? It feels like a very rare case that should not be necessary optimized.

paulomorgado: > @paulomorgado Paulo Morgado FTE Didn't know about the `FormattableString.Invariant($"...")`, thanks!
> But does it seem overcomplicated to think about allocating memory in a location that throws the OutOfMemoryException? It feels like a very rare case that should not be necessary optimized.
> 
> And as I see [here](https://sharplab.io/#v2:D4AQTAjAsAUCAMACEEB0BxANgewEYENMBLAL3wBcjsA7AbllhAGZkxEBhRAb1kT+RYgALIgCyACgCU3Xvzk5qAc0TUApgHd2+AA74AxkXIBPRAF5E8ejDlzyACwBO2dSo2IA8gFdy7gGajVAFtsByMAUQAPPVVtShpxADEQwIpyfFxMVQBlcgciJVQASWoAN3w8/GpycQASACIAJVUAR09VAGdyVQATRD0dfUMTInbEcmxsRExyxVUALm41TQGDYwBfVDrJSSs5Ndg1oA===), `FormattableString.Invariant($"...")` will not protect us from boxing anyway.

I was not implying that it was better, or even good. I was just saying that it was the only way of getting the same outcome out of an interpolated string.

Just use an interpolated string with the `.ToString(CultureInfo.InvariantCulture)` invocation in the holes. If all holes are to be filled with strings, the compiler will choose `string.Concat(string[])` instead of `string.Format`.

hankovich: Got it, done.

src/RecyclableMemoryStream.cs
```diff
@@ -1538,7 +1538,10 @@ private void EnsureCapacity(long newCapacity)
             if (newCapacity > this.memoryManager.MaximumStreamCapacity && this.memoryManager.MaximumStreamCapacity > 0)
             {
                 this.memoryManager.ReportStreamOverCapacity(this.id, this.tag, newCapacity, this.AllocationStack);
-                throw new OutOfMemoryException($"Requested capacity is too large: {newCapacity}. Limit is {this.memoryManager.MaximumStreamCapacity}.");
+
+                var requestedCapacity = newCapacity.ToString(CultureInfo.InvariantCulture);
+                var limit = this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture);
+                throw new OutOfMemoryException($"Requested capacity is too large: {requestedCapacity}. Limit is {limit}.");
```
paulomorgado, line 1544: There's no need to have variables for that.

What's wrong with this?

```csharp
throw new OutOfMemoryException($"Requested capacity is too large: {newCapacity.ToString(CultureInfo.InvariantCulture)}. Limit is {this.memoryManager.MaximumStreamCapacity.ToString(CultureInfo.InvariantCulture)}.");

Other than readability, that is. 😄 
```

hankovich: That's it, it's too long :)
I can inline them if you wish.

paulomorgado: @benmwatson, @grbell-ms, what do you think?

grbell-ms: I don't think we care if we do culture specific formatting or boxing in this rare error case. So let's keep it simple:

```cs
throw new OutOfMemoryException($"Requested capacity is too large: {newCapacity}. Limit is {this.memoryManager.MaximumStreamCapacity}.");
```

src/RecyclableMemoryStream.cs
grbell-ms: Let's simplify a bit:

```csharp
private ArraySegment<byte> GetWritableBuffer(int sizeHint)
{
    this.CheckDisposed();
    if (sizeHint< 0)
    {
        throw new ArgumentOutOfRangeException(nameof(sizeHint), $"{nameof(sizeHint)} must be non-negative.");
    }
    int minimumBufferSize = Math.Max(sizeHint, 1);

    // ...
}
```

src/RecyclableMemoryStream.cs
benmwatson: I'm generally not in favor of introducing temp variables if I don't need them. I don't care about line length in this case. I'm open to culture formatting or not.

benmwatson:
Thanks for doing this impressive amount of standardization!

src/RecyclableMemoryStream.cs
hankovich: > So let's keep it simple

Agreed! Done.

src/RecyclableMemoryStream.cs
hankovich: Sure, done.

grbell-ms (approved):

PR #198, opened 2021-09-02
Avoid enumerator allocations

lbargaoanu:
While small, it's easy to avoid the allocations by using the concrete type instead of `ICollection`.

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -26,6 +26,7 @@ namespace Microsoft.IO.UnitTests
     using System.Buffers;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
     using System.Runtime.InteropServices;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


grbell-ms (approved):

paulomorgado:
Why not just change `ReturnBlocks` to be `internal void ReturnBlocks(List<byte[]> blocks, Guid id, string tag)`? and change the tests to create `List<byte[]>` instances instead of adding allocations to production code?

grbell-ms:
> Why not just change `ReturnBlocks` to be `internal void ReturnBlocks(List<byte[]> blocks, Guid id, string tag)`? and change the tests to create `List<byte[]>` instances instead of adding allocations to production code?

The new allocations are all in test code, and they are pretty tiny.

paulomorgado:
But that shouldn't be a reason for using performance anti-patterns. Should it?

benmwatson:
Sorry for the extreme delay in looking at this PR. I've been out of the office quite a bit lately and have had to just ignore some things. First, thank you @lbargaoanu for bringing up this issue and submitting a PR. A nice, simple fix. I agree that the test allocs aren't that big a deal, but since @paulomorgado has already created an alternative that does the "right" allocations in the test more directly, I'm going to take that version. I think it's been a while since I've published a new version of RMS, so I'll try to do that today.

PR #201, opened 2021-09-16
Add GetStream overrides for ReadOnlyMemory<byte>

### LINKED ISSUE #199: Add GetStream(ReadOnlyMemory<byte>) or replace GetStream(Memory<byte>)
As far as I have understood the documentation, the buffer given using `GetStream(Memory<byte>)` will be copied into the stream. If thats the case, `ReadOnlyMemory<byte>` should be sufficient.

I think this would also help the API to make it more clear that the buffer given to initialize the stream won't be used any further.

I'm stuck with an API that gives me an `ReadOnlyMemory<byte>` buffer which I want to use to initialize the stream.

benmwatson:
Resolves #199 

This adds `GetStream` overrides that accept `ReadOnlyMemory<byte>`, and a couple of unit tests.

I thought about obsoleting the methods that take `Memory<byte>`, but I think there's no compelling reason to do so since the two types are not actually derived from each other, and I don't want to force a conversion, even implicitly, if not necessary.

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -2476,6 +2476,31 @@ namespace Microsoft.IO.UnitTests
             RMSAssert.BuffersAreEqual(buffer.Span, stream.GetBuffer(), buffer.Length);
             Assert.That(buffer, Is.Not.SameAs(stream.GetBuffer()));
         }
+
+        [Test]
+        public void GetStreamWithReadOnlyMemoryBuffer()
+        {
+            var memMgr = this.GetMemoryManager();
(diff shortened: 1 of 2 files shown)


grbell-ms:
Isn't taking a `ReadOnlySpan<byte>` preferable to taking a `ReadOnlyMemory<byte>`?

benmwatson:
Hmm..., maybe I should just obsolete all the overloads that take `Memory`, because all of these types can be brought back down to `Span`, and therefore `ReadOnlySpan`.

grbell-ms (approved):

PR #202, opened 2021-09-17
Avoid allocating a closure in CopyToAsync

lbargaoanu:
The existing code passes less parameters, but a small closure is allocated on every call, so perhaps it's worth to be explicit.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -26,7 +26,6 @@ namespace Microsoft.IO
     using System.Buffers;
     using System.Collections.Generic;
     using System.Diagnostics;
-    using System.Globalization;
     using System.IO;
     using System.Runtime.CompilerServices;
     using System.Threading;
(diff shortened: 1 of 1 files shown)


paulomorgado:
Good catch!

But that doesn't mean it can't be a local function.

src/RecyclableMemoryStream.cs
```diff
@@ -615,6 +598,21 @@ async Task CopyToAsyncImpl(CancellationToken ct, long totalBytesToWrite)
                     return destination.WriteAsync(this.largeBuffer, (int)startPos, (int)count, cancellationToken);
                 }
             }
+            async Task CopyToAsyncImpl(Stream destination, long startPos, long count, CancellationToken cancellationToken)
```
Therzok, line 601: `static` local function could ensure nothing adds a closure capture in the future and expresses intent on not allocating closures

lbargaoanu: It needs `this`. I'd rather not pass it explicitly :)

lbargaoanu:
> 
> 
> Good catch!
> 
> But that doesn't mean it can't be a local function.

It is a local function.

src/RecyclableMemoryStream.cs
lbargaoanu: Perhaps just make it a member function instead?

paulomorgado:
> > But that doesn't mean it can't be a local function.
> 
> It is a local function.

I see it now. But why did you change its scope. It needs only to be scoped to that particular `else`.

lbargaoanu:
For separation. But I think I'll make it a member function to make things even more explicit.

paulomorgado:
> For separation. But I think I'll make it a member function to make things even more explicit.

When you increase the scope, you're not separating. Making it a full member function is even worst.

That method is used only in that `else` and it should be scoped to that `else` only. That makes it "private" to that block. The scope of code artifacts should be the smallest possible to avoid unforeseen uses and side effects from modification.

lbargaoanu:
I see your poin,t but my vote is for a member function. It seems the right compromise here.

paulomorgado:
> I see your point, but my vote is for a member function. It seems the right compromise here.

Why?

lbargaoanu:
https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/pull/202#discussion_r710931819

paulomorgado:
The local function has no compromises. Any compromise would be the wrong compromise.

benmwatson:
What's wrong with just passing those two additional parameters into the local function? That's a smaller change, as @paulomorgado, I don't see the downside here.

lbargaoanu:
It's easy to capture smth again, it's what led to the allocation in the first place. Having some separation from the calling code makes it explicit that we don't want to capture anything.

grbell-ms:
> It's easy to capture smth again, it's what led to the allocation in the first place. Having some separation from the calling code makes it explicit that we don't want to capture anything.

But you can fix that by making the local function static. You just need to pass either `this` or `blocks` and `blockAndOffset` as parameters.

lbargaoanu:
Yes, you can make any member function static by explicit;y passing `this`, but I'd rather not write that code myself :)

paulomorgado:
Too many parameters and some will end up being passed by stack instead of registers.

I agree that it can happen again.

I wonder if [ClrHeapAllocationAnalyzer](https://www.nuget.org/packages/ClrHeapAllocationAnalyzer/) catches this.

grbell-ms:
The codegen differences between a static local function and a method are minimal. The static function has a couple of more instruction of stack shuffling, but it's associated state machine is ~36 bytes smaller.
https://sharplab.io/#gist:e0be6566b653dba8f1da54de00d329a4

benmwatson:
@grbell-ms Are you recommending one particular way or the other? Is it just kind of arbitrary? It would be nice to come to a good agreement or compromise on this PR.

grbell-ms:
@benmwatson  I'm coming down in favor of the static local function. 

- It prevents us from making this mistake in the future
- There's no real perf difference
- We don't pollute the class with methods with only a single reference

paulomorgado:
> @benmwatson Ben Watson FTE I'm coming down in favor of the static local function.
> 
> * It prevents us from making this mistake in the future
> * There's no real perf difference
> * We don't pollute the class with methods with only a single reference

That's also my reasoning.

It requires C# 8.0 or above, but we are already using C# 9.0.

We need to make sure we are running all tests for code targeting .NET Framework.

benmwatson:
@lbargaoanu I see you closed this. Are you going to re-open and do the other way? If not, @paulomorgado or @grbell-ms, would one of you be willing to redo it in the recommended way. I'd love to get this optimization into the next release.

lbargaoanu:
I'd rather not write that code myself. I just don't see the value :)

PR #204, opened 2021-09-17
Start with capacity zero and let EnsureCapacity set it

### LINKED ISSUE #203: Blocks list initial capacity
Another small allocation :)
The initial capacity is [one](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/a2ef5ed3e23161aa1bf98ed0fb840fdc8eb60645/src/RecyclableMemoryStream.cs#L98). Then, in the constructor, the capacity is increased to two ([the default settings](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/a2ef5ed3e23161aa1bf98ed0fb840fdc8eb60645/src/RecyclableMemoryStream.cs#L1559)). This causes an allocation.
Perhaps leave the default (zero in .Net 5) and avoid the allocation? Or if one is hepful in other cases, start with zero just in the default case?

lbargaoanu:
Fixes #203. `Add` is only called by `EnsureCapacity` so I think the capacity will always be set beforehand.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -95,7 +95,7 @@ namespace Microsoft.IO
         /// <summary>
         /// All of these blocks must be the same size.
         /// </summary>
-        private readonly List<byte[]> blocks = new List<byte[]>(1);
+        private readonly List<byte[]> blocks = new List<byte[]>();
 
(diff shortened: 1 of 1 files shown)


benmwatson:
Good catch. Thank you.

PR #237, opened 2022-06-13
Add overloads of GetStream that take a long requestedSize

### LINKED ISSUE #231: RecyclableMemoryStreamManager.GetStream should accept requiredSize as long
The `RecyclableMemoryStream` constructor supports passing a `requestedSize` as `long`, but `RecyclableMemoryStreamManager.GetStream` only accepts `int`. Because of this, if I want a recyclable memory stream that is the same size as another stream, then I have to do a cast:
```
Stream anotherStream;
var recyclable = streamManager.GetStream("tag", (int)anotherStream.Length);
```

benmwatson:
Resolves #231 

This adds overloads to `RecyclableMemoryStreamManager.GetStream` that take `long requestedSize` instead of `int`. This creates consistency between `long RecyclableMemoryStream.Length` and `GetStream` and reduces the need to cast.

### DIFF
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -760,6 +760,17 @@ namespace Microsoft.IO
         /// <returns>A <c>MemoryStream</c>.</returns>
         public MemoryStream GetStream(string tag, int requiredSize)
         {
+            return this.GetStream(tag, (long)requiredSize);
+        }
+
+        /// <summary>
+        /// Retrieve a new <c>MemoryStream</c> object with the given tag and at least the given capacity.
+        /// </summary>
+        /// <param name="tag">A tag which can be used to track the source of the stream.</param>
+        /// <param name="requiredSize">The minimum desired capacity for the stream.</param>
+        /// <returns>A <c>MemoryStream</c>.</returns>
+        public MemoryStream GetStream(string tag, long requiredSize)
+        {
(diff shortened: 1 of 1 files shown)


grbell-ms:
It's a shame to add yet more overloads to `GetStream()`. Might adding just `GetStream(Guid id, string tag, long requiredSize, bool asContiguousBuffer)` be better?

paulomorgado:
> It's a shame to add yet more overloads to `GetStream()`. Might adding just `GetStream(Guid id, string tag, long requiredSize, bool asContiguousBuffer)` be better?

Or go the way of `XmlReader`/`XmlWriter` and add a settings class. Or `in` struct.

benmwatson:
It does seem like a setting struct might be the way to go at this point. That would be a good idea for a new major version revision, and we can clean up the APIs. I'll go with just adding a single new overload for now.

paulomorgado:
A `struct` is a good idea because it will not require allocations.

However, this might grow over time and the copying might become more costly than a short lived `class` allocation.

It can be passed as `in`, but that's not available for every target. it could be replaced with `ref` for those targets, though.

In any way, this would be an added overload that doesn't require an overhaul of the API.

paulomorgado:
A `struct` is a good idea because it will not require allocations.

However, this might grow over time and the copying might become more costly than a short lived `class` allocation.

It can be passed as `in`, but that's not available for every target. it could be replaced with `ref` for those targets, though.

In any way, this would be an added overload that doesn't require a overwalk

PR #240, opened 2022-06-28
use 6.0.301 sdk, add net6 target, drop out of support targets

### LINKED ISSUE #239: Add NET6 target
Current LTS target .NET 6 should be added.
Conversely, .NET 5 and .NET Core 2.1 are out of support and should probably be dropped.

ovska:
- Add net6.0 target, drop net5.0 and netcoreapp2.1

closes #239

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "5.0.403",
+    "version": "6.0.301",
     "rollForward": "latestMinor"
   }
 }
--- a/src/Microsoft.IO.RecyclableMemoryStream.csproj
+++ b/src/Microsoft.IO.RecyclableMemoryStream.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;net462;netcoreapp2.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net462;net6.0</TargetFrameworks>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <RootNamespace>Microsoft.IO</RootNamespace>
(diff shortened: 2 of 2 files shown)


benmwatson:
I'm not in favor of this PR for a few reasons:

1. As a library, it is used in many situations, including in systems that have not yet moved to .NET 6. This change will cause compatibility issues with those.
2. There are no new features being added that rely on .NET 6.
3. Just because .NET 5 is out of support doesn't mean that it can't serve as a minimal compatible version for all versions after.

paulomorgado:
> I'm not in favor of this PR for a few reasons:
> 
> 1. As a library, it is used in many situations, including in systems that have not yet moved to .NET 6. This change will cause compatibility issues with those.
> 2. There are no new features being added that rely on .NET 6.
> 3. Just because .NET 5 is out of support doesn't mean that it can't serve as a minimal compatible version for all versions after.

It can be a bit confusing, but a .NET version is both a runtime version and an API surface version. In the case of NuGet packages, it's always an API surface version.

The major version change implies that there might be breaking changes (and there are some) but if they are not in the APIs used by the library, keeping the lowest API surface increases the number of possible usages.

LeaFrock:
While `netstandard 2.0/2.1` keeps the basic framework support,

* `net462` is the compatible implementation for NFX.
* `netcore2.1` is imported by this [commit](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/commit/86fc159750fccd66e916133a80f8f33df3cbd1f7) which wants to use `Span`-related API surface in `netstandard 2.1`.
* `net5` is imported because of [GC.AllocateUninitializedArray](https://docs.microsoft.com/en-us/dotnet/api/system.gc.allocateuninitializedarray).

Overall, I realize it's reasonable. 

So if `net7` provides new API surface which is useful for specific optimization of `NET_7_0_OR_GREATER`, this lib may add `net7` as a new target framework. It has nothing to do with whether the target framwork is STS or LTS. Do I understand right?

benmwatson:
> So if net7 provides new API surface which is useful for specific optimization of NET_7_0_OR_GREATER, this lib may add net7 as a new target framework. It has nothing to do with whether the target framwork is STS or LTS. Do I understand right?

Correct.

PR #244, opened 2022-08-26
Fix ToString to correctly handle disposed streams.

### LINKED ISSUE #242: RecyclableMemoryStream.ToString should not throw if disposed
From https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=net-6.0
Notes to Inheritors
Your [ToString()](https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=net-6.0#system-object-tostring) override should not throw an exception.

This comes up for us in unit tests that use moq. it remembers all invocations and their parameters and, if a verification failure happens, it will ToString all of the parameters to make the error report but in the case of Stream arguments, if it was a RecyclableMemoryStream that has long since been disposed, the exception from its ToString prevents meaningful diagnostics from moq.

benmwatson:
Resolves #242 

Fix `RecyclableMemoryStream.ToString` to show status of disposable, and some still-valid info after the stream is disposed.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -1087,7 +1087,15 @@ namespace Microsoft.IO
         /// </summary>
         public override string ToString()
         {
-            return $"Id = {this.Id}, Tag = {this.Tag}, Length = {this.Length:N0} bytes";
+            if (!this.disposed)
+            {
+                return $"Id = {this.Id}, Tag = {this.Tag}, Length = {this.Length:N0} bytes";
+            }
+            else
+            {
+                // Avoid properties because of the dispose check, but the fields themselves are not cleared.
+                return $"Disposed: Id = {this.id}, Tag = {this.tag}";
+            }
         }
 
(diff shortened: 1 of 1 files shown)


src/RecyclableMemoryStream.cs
```diff
@@ -1087,7 +1087,15 @@ public override void Write(ReadOnlySpan<byte> source)
...
         public override string ToString()
         {
-            return $"Id = {this.Id}, Tag = {this.Tag}, Length = {this.Length:N0} bytes";
+            if (!this.disposed)
+            {
+                return $"Id = {this.Id}, Tag = {this.Tag}, Length = {this.Length:N0} bytes";
+            }
+            else
+            {
+                // Avoid properties because of the dispose check, but the fields themselves are not cleared.
```
grbell-ms, line 1096: This is also true of the `length` field. We could add the final length or something to the string.

grbell-ms (approved):

PR #265, opened 2023-02-03
Change when pool stats are reported

benmwatson:
Pool stats are currently only reported when blocks/buffers are returned. If you have a ton of long-lived streams being created, this can lead to a very skewed set of statistics. It would be fairer to also update stats when blocks/buffers are retrieved, but this might double the times the API is called.

Instead, I moved the reporting to stream creation and disposal. This should balance things out.

I also changed the method signature to be empty because it was using member variables all the time anyway.

### DIFF
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -555,9 +555,6 @@ namespace Microsoft.IO
             }
 
             Interlocked.Add(ref this.largeBufferInUseSize[poolIndex], -buffer.Length);
-
-            ReportUsageReport(this.smallPoolInUseSize, this.smallPoolFreeSize, this.LargePoolInUseSize,
-                              this.LargePoolFreeSize);
         }
 
         /// <summary>
@@ -599,9 +596,6 @@ namespace Microsoft.IO
                     break;
                 }
             }
-
-            ReportUsageReport(this.smallPoolInUseSize, this.smallPoolFreeSize, this.LargePoolInUseSize,
-                              this.LargePoolFreeSize);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


grbell-ms (approved):
This should be both more efficient and complete. 🙂

src/RecyclableMemoryStreamManager.cs
```diff
@@ -707,9 +698,9 @@ internal void ReportStreamOverCapacity(Guid id, string tag, long requestedCapaci
             this.StreamOverCapacity?.Invoke(this, new StreamOverCapacityEventArgs(id, tag, requestedCapacity, this.MaximumStreamCapacity, allocationStack));
         }
 
-        internal void ReportUsageReport(long smallPoolInUseBytes, long smallPoolFreeBytes, long largePoolInUseBytes, long largePoolFreeBytes)
+        internal void ReportUsageReport()
         {
-            this.UsageReport?.Invoke(this, new UsageReportEventArgs(smallPoolInUseBytes, smallPoolFreeBytes, largePoolInUseBytes, largePoolFreeBytes));
+            this.UsageReport?.Invoke(this, new UsageReportEventArgs(this.smallPoolInUseSize, this.smallPoolFreeSize, this.LargePoolInUseSize, this.LargePoolFreeSize));
```
yihpink, line 703: worth handling exceptions? usage reporting is not the primary purpose of the manager

yihpink (approved):

paulomorgado (approved):

src/RecyclableMemoryStreamManager.cs
benmwatson: Exception handling could impact the performance of this method by making it non-inlinable, so I'd rather not without some stronger evidence that it should be done here rather than in the user's code.

PR #266, opened 2023-02-04
NET7 and small cleanup

sungam3r:
If some changes seem superfluous, I will revert them. In fact, there are not many changes at all and they are cosmetic.

### DIFF
--- a/src/Microsoft.IO.RecyclableMemoryStream.csproj
+++ b/src/Microsoft.IO.RecyclableMemoryStream.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;netstandard2.1;net462;netcoreapp2.1;net5.0</TargetFrameworks>
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <RootNamespace>Microsoft.IO</RootNamespace>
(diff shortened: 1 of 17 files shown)


paulomorgado:
There are quite a few style and source text changes.

Please, back them up with a .editorconfig file.

sungam3r:
What exactly would you like me to revert? Switch expressions?

src/Microsoft.IO.RecyclableMemoryStream.csproj
```diff
@@ -50,11 +51,11 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All"/>
     <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0'" />
   </ItemGroup>
-  
+
   <ItemGroup>
-    <None Include="..\README.md" Pack="true" PackagePath="\" />
+    <None Include="..\README.md" Pack="true" PackagePath="\" Visible="false" />
```
benmwatson, line 56: Why make this invisible?

benmwatson:
> What exactly would you like me to revert? Switch expressions? Using statements? Both?

He's not saying to revert, just to add an .editorconfig to enforce these style decisions automatically. I am neutral on the specific style used, but I think the .editorconfig is a good idea to enforce style decisions going forward.

Overall, no issues with PR. Just a minor question about why you want to hide README.md in the project. I appreciate having it visible in VS.

src/Microsoft.IO.RecyclableMemoryStream.csproj
sungam3r: Already visible in solution items.

sungam3r:
> Please, back them up with a .editorconfig file.

I misread it. Done.

PR #268, opened 2023-02-06
Add tracking for stream lifetime

benmwatson:
An insufficiently managed pool is indistinguishable from a memory leak. One area that would be nice to have is the ability to track the length of time streams are active. This would inform pool configuration, especially as it changes over time.

In our own service, I would hook this up to a performance counter to track over time.

I've also put the lifetime value in the Dispose ETW event, which is a verbose event.

### DIFF
--- a/src/EventArgs.cs
+++ b/src/EventArgs.cs
@@ -71,16 +71,23 @@
             public string DisposeStack { get; }
 
             /// <summary>
+            /// Lifetime of the stream
+            /// </summary>
+            public TimeSpan Lifetime { get; }
+
+            /// <summary>
             /// Initializes a new instance of the <see cref="StreamDisposedEventArgs"/> class.
             /// </summary>
             /// <param name="guid">Unique ID of the stream.</param>
             /// <param name="tag">Tag of the stream.</param>
+            /// <param name="lifetime">Lifetime of the stream</param>
             /// <param name="allocationStack">Stack of original allocation.</param>
             /// <param name="disposeStack">Dispose stack.</param>
-            public StreamDisposedEventArgs(Guid guid, string tag, string allocationStack, string disposeStack)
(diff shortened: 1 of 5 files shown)


paulomorgado (approved):

src/RecyclableMemoryStream.cs
```diff
@@ -308,13 +311,14 @@ protected override void Dispose(bool disposing)
             }
 
             this.disposed = true;
+            var lifetime = TimeSpan.FromTicks(Stopwatch.GetTimestamp() - this.creationTimestamp);
```
grbell-ms, line 314: This is only correct when the duration of a tick and the timestamp are the same. This is true on Windows now, but it may not be on other platforms or in the future. See the [discussion of why Stopwatch.GetElapsedTime](https://github.com/dotnet/runtime/issues/65858#issuecomment-1050836263) was added for the right way to do it:

```suggestion
            var lifetime = TimeSpan.FromTicks((long)((Stopwatch.GetTimestamp() - this.creationTimestamp) * (TimeSpan.TicksPerSecond / Stopwatch.Frequency)));
```

grbell-ms: On [Unix](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Stopwatch.Unix.cs,10), for example, the smallest increment of a timestamp is only 1 nanosecond! The time reported would then be 100 times too long.

grbell-ms (approved):

src/EventArgs.cs
```diff
@@ -70,17 +70,38 @@ public sealed class StreamDisposedEventArgs : EventArgs
             /// </summary>
             public string DisposeStack { get; }
 
+            /// <summary>
+            /// Lifetime of the stream
```
sungam3r, line 74:
```suggestion
            /// Lifetime of the stream.
```

PR #270, opened 2023-02-07
Add possibility to zero out memory before returning it to the pool

### LINKED ISSUE #267: FeatureRequest: Add opt-in possibility to zeroed-out buffers
Hi together,

Would it be possible to extend the RecyclableMemoryStreamManager by a configuration option to zero-out buffers before returning them to the pool?
I understand that for performance reasons, the buffers are not ever pre-initialized or zeroed-out by default, but i think it would be nice to have the option to configure it if one is willing to take the performance hit.

I could create a PR for this if you think it would be a useful addition.

Best regards,
Lars

lklein53:
Closes #267 
TODO

- ~~Add test cases~~

### DIFF
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -387,7 +387,7 @@ namespace Microsoft.IO
                 // We'll add this back to the pool when the stream is disposed
                 // (unless our free pool is too large)
 #if NET5_0_OR_GREATER
-                block = GC.AllocateUninitializedArray<byte>(this.BlockSize);
+                block = this.ZeroOutBuffer ? GC.AllocateArray<byte>(BlockSize) : GC.AllocateUninitializedArray<byte>(this.BlockSize);
 #else
                 block = new byte[this.BlockSize];
 #endif
@@ -430,7 +430,7 @@ namespace Microsoft.IO
             {
                 if (!this.largePools[poolIndex].TryPop(out buffer))
                 {
-                    buffer = AllocateArray(requiredSize);
+                    buffer = AllocateArray(requiredSize, this.ZeroOutBuffer);
                     createdNew = true;
                 }
                 else
@@ -447,7 +447,7 @@ namespace Microsoft.IO
                 poolIndex = this.largeBufferInUseSize.Length - 1;
 
                 // We still want to round up to reduce heap fragmentation.
-                buffer = AllocateArray(requiredSize);
+                buffer = AllocateArray(requiredSize, this.ZeroOutBuffer);
                 if (this.GenerateCallStacks)
                 {
                     // Grab the stack -- we want to know who requires such large buffers
@@ -466,9 +466,9 @@ namespace Microsoft.IO
             return buffer;
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            static byte[] AllocateArray(long requiredSize) =>
+            static byte[] AllocateArray(long requiredSize, bool zeroInitializeArray) =>
 #if NET5_0_OR_GREATER
-                GC.AllocateUninitializedArray<byte>((int)requiredSize);
+                zeroInitializeArray ? GC.AllocateArray<byte>((int)requiredSize) : GC.AllocateUninitializedArray<byte>((int)requiredSize);
 #else
                 new byte[requiredSize];
 #endif
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -369,6 +369,12 @@ namespace Microsoft.IO
         public bool ThrowExceptionOnToArray { get; set; }
 
         /// <summary>
+        /// Zero out buffers before returning them to the pool.
+        /// </summary>
+        /// <remarks>Setting this to true causes a performance hit and should only be set if one wants to avoid accidental data leaks.</remarks>
+        public bool ZeroOutBuffer { get; set; }
+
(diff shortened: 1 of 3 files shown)


benmwatson:
@lklein53 Are you still working on this PR? I've left a comment, and it needs test cases.

lklein53:
> @lklein53 Are you still working on this PR? I've left a comment, and it needs test cases.

Sorry completly forgot about this. Will pick it up in 2 weeks once i am back from vacation. However i can't see the comment you mentioned.

src/RecyclableMemoryStreamManager.cs
```diff
@@ -368,6 +368,12 @@ public long LargeBuffersFree
         /// set and <see cref="RecyclableMemoryStream.ToArray"/> is called, a <c>NotSupportedException</c> will be thrown.</remarks>
         public bool ThrowExceptionOnToArray { get; set; }
 
+        /// <summary>
+        /// Zero out buffers before returning them to the pool.
+        /// </summary>
+        /// <remarks>Setting this to true causes a performance hit and should only be set if one wants to avoid accidental data leaks.</remarks>
```
grbell-ms, line 374: This flag does more than the comment suggests. It zeros newly allocated buffers, as well as returned buffers.

src/RecyclableMemoryStreamManager.cs
```diff
@@ -632,6 +639,12 @@ internal void ReturnBlock(byte[] block, Guid id, string tag)
             }
         }
 
+        private void ZeroOutMemoryIfEnabled(Span<byte> buffer) {
+            if (this.ZeroOutBuffer) {
+                buffer.Clear();
+            }
+        }
```
grbell-ms, line 646: We don't want to construct a span when we don't have to.

```suggestion
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        private void ZeroOutMemoryIfEnabled(byte[] buffer) {
            if (this.ZeroOutBuffer) {
                Array.Clear(buffer);
            }
        }
```

lklein53: Thanks. Chankged the code

src/RecyclableMemoryStreamManager.cs
lklein53: Thanks. Adjusted the comment

lklein53: Thanks. Adjusted the comment

src/RecyclableMemoryStreamManager.cs
```diff
@@ -460,9 +466,9 @@ internal byte[] GetLargeBuffer(long requiredSize, Guid id, string tag)
             return buffer;
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            static byte[] AllocateArray(long requiredSize) =>
+            static byte[] AllocateArray(long requiredSize, bool zeroInitializeArray) =>
 #if NET5_0_OR_GREATER
-                GC.AllocateUninitializedArray<byte>((int)requiredSize);
+                zeroInitializeArray ? GC.AllocateArray<byte>((int)requiredSize) : GC.AllocateUninitializedArray<byte>((int)requiredSize);
 #else
                 new byte[requiredSize];
```
benmwatson, line 473: What about the < .NET 5.0 case?

benmwatson (changes requested):

benmwatson (approved):

lklein53:
It seems like the pipeline failed with the following message

> [error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.

I haven't found a way to manually trigger the CI Pipeline again. Could you tell me how to do this?

benmwatson:
> I haven't found a way to manually trigger the CI Pipeline again. Could you tell me how to do this?

I'm able to re-run the pipeline, but it hasn't helped. I thought there could be some broken unit tests, but they all run fine on my local machine form your repo.

I wonder if it's a problem with some tests that allocate a ton of memory, and maybe something in the new functionality is causing the test process to crash or exceed some resource constraints.

I can't figure out how to get the logs it mentions. I'll keep digging at it.

lklein53:
> > I haven't found a way to manually trigger the CI Pipeline again. Could you tell me how to do this?
> 
> I'm able to re-run the pipeline, but it hasn't helped. I thought there could be some broken unit tests, but they all run fine on my local machine form your repo.
> 
> I wonder if it's a problem with some tests that allocate a ton of memory, and maybe something in the new functionality is causing the test process to crash or exceed some resource constraints.
> 
> I can't figure out how to get the logs it mentions. I'll keep digging at it.

Thanks a lot. Yeah i also tried them locally and they run fine.

benmwatson:
I've done some digging on this and have discovered the following:

* The test suite will pass if I disable your new test configuration.
* The test suite will pass if I disable all the VeryLargeStream_* tests. 
* Monitoring the test suites on my dev machines shows that the new test configuration uses more memory (~26GB) than the others (~20GB).
* I'm not sure if I can get access to larger runners. I'll have to figure out who to ask on the Microsoft account about this.
* I'm assuming there's something about zeroing out the memory that is causing more memory to actually be committed. In many of the tests, I don't think we're actually explicitly touching all the buffers, so this could be very possible.

My immediate recommendation is that you add a switch to the tests that will skip running the VeryLargeStream* tests in your test configuration.

lklein53:
Thanks for diggin into it. I went with your suggestion

benmwatson (approved):

PR #273, opened 2023-02-12
More optimized remainder calculation

lechu445:
Instead of division then remainder operation, used Math.DivRem which handles it in a more optimized way. It uses only a single division, multiplication, and subtraction instead of two divisions. Moreover, in the future, it can be even more optimized using intrinsics.

See:
https://github.com/dotnet/runtime/issues/5213#issuecomment-260448104

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -1120,14 +1120,14 @@ namespace Microsoft.IO
             {
                 var blockSize = this.memoryManager.BlockSize;
 
-                var block = (int)(this.position / blockSize);
+                var block = (int)Math.DivRem(this.position, blockSize, out var index);
 
                 if (block >= this.blocks.Count)
                 {
                     this.EnsureCapacity(end);
                 }
 
-                this.blocks[block][this.position % blockSize] = value;
+                this.blocks[block][index] = value;
             }
             else
             {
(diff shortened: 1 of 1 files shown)


paulomorgado:
It would be nice if this PR was acompanied by benchmarks.

lechu445:
before:
```
|    Method |     Mean |   Error |   StdDev |
|---------- |---------:|--------:|---------:|
| WriteByte | 184.0 ms | 3.64 ms |  5.34 ms |
| WriteSpan | 366.3 ms | 7.07 ms | 12.00 ms |
```

after:
```
|    Method |     Mean |   Error |   StdDev |
|---------- |---------:|--------:|---------:|
| WriteByte | 115.5 ms | 2.30 ms |  3.23 ms |
| WriteSpan | 319.7 ms | 6.28 ms | 11.49 ms | 
```

code:
```c#
public class Test
{
    private RecyclableMemoryStream subject;
    private readonly byte[] bytes = new byte[] { 1, 2, 3, 4, 5, 6 };

    [IterationSetup]
    public void IterationSetup()
    {
        var manager = new RecyclableMemoryStreamManager();
        this.subject = new RecyclableMemoryStream(manager);
    }

    [Benchmark]
    public void WriteByte()
    {
        for (int i = 0; i < 10_000_000; i++)
        {
            this.subject.WriteByte(1);
        }
    }

    [Benchmark]
    public void WriteSpan()
    {
        for (int i = 0; i < 10_000_000; i++)
        {
            this.subject.Write(bytes.AsSpan());
        }
    }
}
```

paulomorgado:
I was thinking more in the lines of:

```csharp
public class Benchmarks
{
	[Benchmark(Baseline = true)]
	public (long quocient, long remainder) DivAndMod() => (4000L / 1024L, 4000L % 1024L);

	[Benchmark]
	public (long quocient, long remainder) MathDivRem() => Math.DivRem(4000L, 1024L);
}
```

|     Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|----------- |----------:|----------:|----------:|------:|--------:|
|  DivAndMod | 0.2267 ns | 0.0092 ns | 0.0077 ns |  1.00 |    0.00 |
| MathDivRem | 0.0607 ns | 0.0221 ns | 0.0271 ns |  0.25 |    0.12 |

paulomorgado (approved):

grbell-ms (approved):

PR #274, opened 2023-02-13
Enable nullable annotations and warnings

### LINKED ISSUE #254: Enable nullable annotations and warnings
Apply C# 8.0's _Nullable Reference Types_ by setting `<Nullable>enable</Nullable>` and annotating the entire library with nullable reference types.
As of `v2.2.1`, `LangVersion` is already set to `9.0`.

This change may warrant an increment of the major version number:
Enabling nullable annotations is basically emitting and applying both `System.Runtime.CompilerServices.NullableAttribute` and `System.Runtime.CompilerServices.NullableContextAttribute` which, I believe, should be binary compatible. However, if a consumer has either `<Nullable>warnings</Nullable>` set or is using `#nullable enable warnings` at the consuming code lines while also having `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>`, this would be a source breaking change.
Perhaps to consider taking in other _major_ changes, such as #165.

lechu445:
Followed https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/api-guidelines/nullability.md

Resolves #254

### DIFF
--- a/src/EventArgs.cs
+++ b/src/EventArgs.cs
@@ -17,7 +17,7 @@
             /// <summary>
             /// Optional Tag for the event.
             /// </summary>
-            public string Tag { get; }
+            public string? Tag { get; }
 
             /// <summary>
             /// Requested stream size.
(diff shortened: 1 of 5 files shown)


benmwatson:
This is a potential breaking change, so will wait until 3.0 is ready to be worked on. (Hopefully soon.)

sungam3r:
What does _potential breaking_ mean?

lechu445:
It means that even if this change is not a binary breaking change, it can cause compilation errors in projects that have enabled [TreatWarningsAsErrors](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#treatwarningsaserrors).

paulomorgado:
Those warnings mean potential errors and should be addressed.

But I agree that should be on o a major release.

benmwatson:
@lechu445, could you please resolve the conflicts?

lechu445:
@benmwatson done

benmwatson (approved):

PR #275, opened 2023-02-13
Throw ArgumentNullException instead of NullReferenceException in RecyclableMemoryStream ctor

lechu445:
ArgumentNullException is better because says which parameter was null. Throwing NullReferenceException in libraries is a bad practice.


benmwatson:
Given that `RecyclableMemoryStream`'s constructor is internal and can only be called by an instance of `RecyclableMemoryStreamManager`, how is it possible to hit this code (other than a unit test that can cheat)?

lechu445:
You are right.
I am closing it.

PR #278, opened 2023-02-17
Override CopyTo to improve performance

lechu445:
If CopyTo is not overridden, it is used the default implementation in Stream class which is inefficient as it uses ArrayPool and writes in loop.

### DIFF
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -555,7 +555,7 @@ namespace Microsoft.IO
         /// <inheritdoc/>
         public override void CopyTo(Stream destination, int bufferSize)
         {
-            CopyToAsync(destination, bufferSize, CancellationToken.None).GetAwaiter().GetResult();
+            WriteTo(destination, this.position, this.length - this.position);
         }
 #endif
 
--- a/src/RecyclableMemoryStream.cs
+++ b/src/RecyclableMemoryStream.cs
@@ -551,6 +551,14 @@ namespace Microsoft.IO
             return this.largeBuffer;
         }
 
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
+        /// <inheritdoc/>
+        public override void CopyTo(Stream destination, int bufferSize)
+        {
+            CopyToAsync(destination, bufferSize, CancellationToken.None).GetAwaiter().GetResult();
+        }
+#endif
+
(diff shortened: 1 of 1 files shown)


lechu445:
before:
```
|    Method |     Mean |     Error |    StdDev | Allocated |
|---------- |---------:|----------:|----------:|----------:|
| WriteSpan | 1.508 us | 0.0459 us | 0.1272 us |     968 B |
```

after:
```
|    Method |     Mean |     Error |    StdDev |   Median | Allocated |
|---------- |---------:|----------:|----------:|---------:|----------:|
| WriteSpan | 1.042 us | 0.0266 us | 0.0715 us | 1.000 us |     632 B |
```

benchmark code:
```c#
[MemoryDiagnoser]
public class Test
{
    private RecyclableMemoryStream subject;
    private MemoryStream dest;
    private readonly byte[] bytes = new byte[] { 1, 2, 3, 4, 5, 6 };

    [IterationSetup]
    public void IterationSetup()
    {
        var manager = new RecyclableMemoryStreamManager();
        this.subject = new RecyclableMemoryStream(manager);
        for (int i = 0; i < 1000; i++)
        {
            this.subject.Write(bytes.AsSpan());
        }
        this.dest = new MemoryStream(10_000);
    }

    [Benchmark]
    public void WriteSpan() => this.subject.CopyTo(this.dest);
}
```

src/RecyclableMemoryStream.cs
```diff
@@ -551,6 +551,14 @@ public override byte[] GetBuffer()
             return this.largeBuffer;
         }
 
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
+        /// <inheritdoc/>
+        public override void CopyTo(Stream destination, int bufferSize)
+        {
+            CopyToAsync(destination, bufferSize, CancellationToken.None).GetAwaiter().GetResult();
```
grbell-ms, line 558: Sync over async is an antipattern. Let's reuse `WriteTo`.
```suggestion
            WriteTo(destination, this.position, this.length - this.position);
```

grbell-ms (changes requested):

lechu445:
after applying the suggestion, the performance is even better
```
|    Method |     Mean |    Error |   StdDev | Allocated |
|---------- |---------:|---------:|---------:|----------:|
| WriteSpan | 672.2 ns | 32.72 ns | 91.22 ns |     632 B | 
```

benmwatson (approved):

Issue #256, opened 2022-12-08
Reduce the number of Method/Constructor overloads.


grbell-ms:
- Remove obsolete overloads (e.g ` GetStream(Memory<byte> buffer)` et al.)
- Remove redundant overloads (e.g. no need for an `int` overload when there's an equivalent `long` overload)
- Consider creating a `MemoryStreamManagerOptions` struct to simplify creation and DI integration.

lechu445:
API proposal:
```c#
    /// <summary>
    /// Provides the ability for the user to define custom behavior when using RecyclableMemoryStream.
    /// </summary>
    public struct MemoryStreamManagerOptions
    {
        /// <summary>
        /// Size of each block that is pooled. Must be > 0.
        /// </summary>
        /// <remarks>
        /// By default, it's set to 131072.
        /// </remarks>
        public int BlockSize { get; set; }

        /// <summary>
        /// Each large buffer will be a multiple/exponential of this value.
        /// </summary>
        /// <remarks>
        /// By default, it's set to 1048576.
        /// </remarks>
        public int LargeBufferMultiple { get; set; }

        /// <summary>
        /// Buffers larger than this are not pooled.
        /// </summary>
        /// <remarks>
        /// By default, it's set to 134217728.
        /// </remarks>
        public int MaximumBufferSize { get; set; }

        /// <summary>
        /// Switch to exponential large buffer allocation strategy.
        /// </summary>
        /// <remarks>
        /// By default, it's set to false.
        /// </remarks>
        public bool UseExponentialLargeBuffer { get; set; }

        /// <summary>
        /// Maximum number of bytes to keep available in the small pool before future buffers.
        /// </summary>
        /// <remarks>
        /// By default, it's set to 0.
        /// </remarks>
        public long MaximumSmallPoolFreeBytes { get; set; }

        /// <summary>
        /// Maximum number of bytes to keep available in the large pool before future buffers get dropped for garbage collection.
        /// </summary>
        /// <remarks>
        /// By default, it's set to 0.
        /// </remarks>
        public long MaximumLargePoolFreeBytes { get; set; }
    }
```
```c#
    public partial class RecyclableMemoryStreamManager
    {
        /// <summary>
        /// Initializes the memory manager with the default block/buffer specifications and maximum free bytes specifications.
        /// </summary>
        /// <param name="options">Defines the customized behavior of the <see cref="RecyclableMemoryStream"/>.</param>
[...]

PR #279, opened 2023-02-17
Remove obsolete overloads

lechu445:
Contributes to #256

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -2755,33 +2755,6 @@ namespace Microsoft.IO.UnitTests
         }
 
         [Test]
-        [Obsolete("GetStream(Memory<byte>) is obsolete.")]
-        public void GetStreamWithMemoryBuffer()
-        {
-            var memMgr = this.GetMemoryManager();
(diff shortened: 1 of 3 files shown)


benmwatson:
This is fine, but will be postponed until 3.0 is a little closer to being worked on.

paulomorgado:
The [overload of the Obsolete attribute being used](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute.-ctor#system-obsoleteattribute-ctor(system-string)) leads to a value of **false** for [`IsError`](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute.iserror).

It is a good practice to not remove obsolete API without preventing it's use first by causing compiler error by using [the constructor that allows to specify `IsError`](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute.-ctor#system-obsoleteattribute-ctor(system-string-system-boolean)) with **true** to cause compile-time errors and keep the binary compability.

benmwatson (approved):

PR #280, opened 2023-02-17
Remove redundant int overloads

lechu445:
Contributes #256


paulomorgado:
This should have an obsolescence plan starting with `ObsoleteAttribute.IsError` **false** and then **true**

benmwatson (approved):

PR #286, opened 2023-03-04
Add continuous benchmarking

### LINKED ISSUE #277: Continous benchmarking
As this library is performance oriented, I think benchmark tests could be useful to spot any perf regressions. Also performance results can motivate contributors to improve the results.

As a first step, I'd suggest to use [BenchmarkDotNet](https://benchmarkdotnet.org) for the tests.

And in second step prepare Github action using [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) to enable Continous Benchmarking.

lechu445:
resolves #277 

What is needed to configure:
- create remote branch gh-pages
- in `Settings -> Pages -> Build and deployment` select branch gh-pages
- the page with benchmark results will be accessible in https://microsoft.github.io/Microsoft.IO.RecyclableMemoryStream/dev/bench/

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,35 @@
+name: Benchmark Tests
+on:
+  push:
+    branches:
+      - master
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  benchmark:
+    name: Run Benchmark.Net
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


lechu445:
here is how it looks for my fork https://lechu445.github.io/Microsoft.IO.RecyclableMemoryStream/dev/bench/

benmwatson (approved):

PR #313, opened 2023-11-20
Change MaximumFreeLargePoolBytes to be a global limit

### LINKED ISSUE #290: Large pool limit should apply to whole pool, not each slot
Right now, the large pool limit is tested per slot. This is unintuitive and probably wrong, and I no longer remember the reason it was originally implemented this way. There should be a global limit.

Having a global limit will be more intuitive to configure, as well as allowing each slot in the pool to fill up more ideally with what is being used.

benmwatson:
This changes the internal meaning of `MaximumFreeLargePoolBytes` to be a global limit to the pool size, similar to the limit for the block pool.

Up to now, if you had set a 64 MB limit, for example, this would apply for each of the large pools created. For example, if you specified a MaximumBufferSize of 64 MB and a LargePoolMultiple of 8MB, then you would have 8 sub-pools, and each one would have a limit of 64 MB. 

This behavior was not publicly documented in the interface, and could lead to higher memory usage than expected.

I believe the original reason for this was to ensure that the different sub-pools didn't starve requests for specific buffer sizes. This change might introduce this problem, thus there is another implementation tweak: if the target sub-pool is empty, it will move on until a free buffer is found. If no free buffer is found, then the original target size is used to allocate a new buffer, which then has a chance to be pooled.

Resolves #290

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -1,4 +1,4 @@
-﻿// ---------------------------------------------------------------------
+// ---------------------------------------------------------------------
 // Copyright (c) 2015 Microsoft
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


benmwatson:
Decided not to do this after all. Too many other changes that could have a negative performance impact.

PR #315, opened 2023-11-21
Introduce an Options class for managing RMSManager settings

### LINKED ISSUE #256: Reduce the number of Method/Constructor overloads.

benmwatson:
This PR creates an `Options` struct to contain all the initial `RecyclableMemoryStreamManager` settings. 

Most of the code changes are just simple refactoring.

A few other things of note:

* I removed most Manager constructors.
* I changed the exception thrown when validating the configuration to be `InvalidOperationException` since it feels like the mistake happens at a remove from the direct constructor call. If you disagree, please yell. I could go either way on this. I'm not sure what the standard approach for this pattern is.
* I added a couple of new tests to improve code coverage of other methods that have been added recently without corresponding tests.
* I fixed some spelling issues in comments, but I decided to stop doing this. I'll deal with nitpicky details like this in a separate PR.

Resolves #256

### DIFF
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -155,7 +155,7 @@ namespace Microsoft.IO
         /// <summary>
         /// Parameters for customizing the behavior of <see cref="RecyclableMemoryStreamManager"/>
         /// </summary>
-        public struct Options
+        public class Options
         {
             /// <summary>
             /// Gets or sets the size of the pooled blocks. This must be greater than 0.
--- a/src/RecyclableMemoryStreamManager.cs
+++ b/src/RecyclableMemoryStreamManager.cs
@@ -80,7 +80,7 @@ namespace Microsoft.IO
         private long smallPoolFreeSize;
         private long smallPoolInUseSize;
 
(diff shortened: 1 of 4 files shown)


src/RecyclableMemoryStreamManager.cs
```diff
@@ -319,83 +153,215 @@ public long LargeBuffersFree
         }
 
         /// <summary>
-        /// How many bytes of small free blocks to allow before we start dropping
-        /// those returned to us.
+        /// Parameters for customizing the behavior of <see cref="RecyclableMemoryStreamManager"/>
+        /// </summary>
+        public struct Options
```
grbell-ms, line 158: We might want to make `Options` a class so it will play nice with `IOptions` which is defined as `public interface IOptions<out TOptions> where TOptions : class`

grbell-ms (approved):

PR #338, opened 2024-03-18
Add important note to avoid memory leak by wrong usage of the lib

AndrePostiga:
Use [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) was very important to avoid memory leaks on my implementations, but I note that the incorrect usage/instanciation can make an memory leaks in gen2 memory.

To avoid the incorrect usage I've added an enphasys on usage note to avoid wrong implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ class Program
     }
 }
 ```
-| **_IMPORTANT_** | Note that `RecyclableMemoryStreamManager` should be declared once and it will live for the entire process lifetime. It is perfectly fine to use multiple pools if you desire, especially if you want to configure them differently..|
+| **_IMPORTANT_** | Note that `RecyclableMemoryStreamManager` should be declared once and it will live for the entire process lifetime. It is perfectly fine to use multiple pools if you desire, especially if you want to configure them differently.|
 |-|:-|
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,8 +103,8 @@ class Program
     }
 }
 ```
-
-Note that `RecyclableMemoryStreamManager` should be declared once and it will live for the entire process lifetime. It is perfectly fine to use multiple pools if you desire, especially if you want to configure them differently.
+| **_IMPORTANT_** | Note that `RecyclableMemoryStreamManager` should be declared once and it will live for the entire process lifetime. It is perfectly fine to use multiple pools if you desire, especially if you want to configure them differently..|
+|-|:-|
```
sungam3r, line 107:
```suggestion

⚠️ Note that `RecyclableMemoryStreamManager` should be declared once and it will live for the entire process lifetime. It is perfectly fine to use multiple pools if you desire, especially if you want to configure them differently.
```

AndrePostiga: I think this note is too important for a single moji 😢, but it's better than the current description.
What do you think?

sungam3r: Up to you. Just remove doubled dot at the end.

AndrePostiga: thx :)
done

benmwatson (approved):

PR #352, opened 2024-08-23
Utilize .NET 8.0 trimming-safe EventSourcePrimitives for logging

### LINKED ISSUE #349: Missing `[RequiresUnreferencedCodeAttribute]` annotation for trimming
Greetings,

due to the call of `this.WriteEvent`, the library throws trim warnings since `RequiresUnreferencedCodeAttribute` is not currently propagated to the invoking methods.

# Minimal reproduction

1. Create new console app: `dotnet new console`
2. Add RecycleableMemoryStream package: `dotnet add package Microsoft.IO.RecyclableMemoryStream`
3. Replace contents of `Program.cs` with: `_ = new Microsoft.IO.RecyclableMemoryStreamManager();`
4. Publish trimmed assembly, and show all trim warnings: `dotnet publish -p:PublishTrimmed=true -p:EnableTrimAnalyzer=true -p:TrimmerSingleWarn=false`

# Expected result

The compilation and trimming concludes without any warnings

# Actual result

The compilation emits trim warnings such as: `/_/src/Events.cs(83,21): Trim analysis warning IL2026: Microsoft.IO.RecyclableMemoryStreamManager.Events.MemoryStreamCreated(Guid, String, Int64, Int64): Using member 'System.Diagnostics.Tracing.EventSource.WriteEvent(Int32, Object[])' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. EventSource will serialize the whole object graph. Trimmer will not safely handle this case because properties may be trimmed.
/_/src/Events.cs(235,21): Trim analysis warning IL2026: Microsoft.IO.RecyclableMemoryStreamManager.Events.MemoryStreamDiscardBuffer(Guid, String, RecyclableMemoryStreamManager.Events.MemoryStreamBufferType, RecyclableMemoryStreamManager.Events.MemoryStreamDiscardReason, Int64, Int64, Int64, Int64, Int64, Int64): Using member 'System.Diagnostics.Tracing.EventSource.WriteEvent(Int32, Object[])' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. EventSource will serialize the whole object graph. Trimmer will not safely handle this case because properties may be trimmed.`

Adding `[RequiresUnreferencedCodeAttribute]` to all methods calling `this.WriteEvent` would properly forward the trim warnings, not causing the library to produce trim warnings.

zeroskyx:
This PR fixes #349 by utilizing .NET 8.0 trimming-safe EventSourcePrimitives for logging.

Please note that this requires .NET 8.0 which might need to be added to the build pipeline.

### DIFF
--- a/src/Events.cs
+++ b/src/Events.cs
@@ -80,7 +80,11 @@ namespace Microsoft.IO
             {
                 if (this.IsEnabled(EventLevel.Verbose, EventKeywords.None))
                 {
+#if NET8_0_OR_GREATER
+                    this.WriteEvent(1, new EventSource.EventSourcePrimitive[] { guid, tag ?? string.Empty, requestedSize, actualSize });
+#else
                     this.WriteEvent(1, guid, tag ?? string.Empty, requestedSize, actualSize);
+#endif
                 }
             }
 
@@ -97,7 +101,11 @@ namespace Microsoft.IO
             {
                 if (this.IsEnabled(EventLevel.Verbose, EventKeywords.None))
                 {
+#if NET8_0_OR_GREATER
+                    this.WriteEvent(2, new EventSource.EventSourcePrimitive[] { guid, tag ?? string.Empty, lifetimeMs, allocationStack ?? string.Empty, disposeStack ?? string.Empty });
+#else
                     this.WriteEvent(2, guid, tag ?? string.Empty, lifetimeMs, allocationStack ?? string.Empty, disposeStack ?? string.Empty);
+#endif
                 }
             }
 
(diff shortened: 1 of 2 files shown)


zeroskyx:
@microsoft-github-policy-service agree

maxkatz6 (approved):
This PR fixes the trimming warning.

Previously, this code was using `WriteEvent(int eventId, params object[] args)` overload, which was involving potentially trimmer unsafe object serialization and is marked `RequiresUnreferencedCode` because of that.
Unsafe object serialization is not really the case in this code, every argument is a primitive. But compiler only sees that `RequiresUnreferencedCode` method is used and emits warning.

With this PR, new `WriteEvent(int eventId, params EventSourcePrimitive[] args)` overload is used. It is safe and is not marked with `RequiresUnreferencedCode` attribute, because each [EventSourcePrimitive](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource.eventsourceprimitive?view=net-10.0) can only be created from the primitive by implicit cast.

benmwatson:
Replaced by #443

PR #355, opened 2024-09-03
Refactor buffer management for ArrayPool

paulomorgado:
This is mostly a provocatory PR.

Using `ArrayPool<byte>` for this class has been mentioned around the interwebs.

I'm not sure a drop-in using `ArrayPool<byte>` is possible, or even a good idea, but here it is.

- Added [Ignore("ArrayPool")] to tests in Tests.cs to skip tests reliant on ArrayPool.
- Commented out ArrayPool-related assertions in Tests.cs.
- Modified buffer length check in ReturnTempBuffer method in RecyclableMemoryStream.cs.
- Replaced custom small pool with ArrayPool<byte> in RecyclableMemoryStreamManager.cs.
- Removed small pool fields and related public properties/methods.
- Updated GetBlock and ReturnBlock methods to use ArrayPool<byte>.
- Removed small pool state reporting methods.
- Added ArrayPool<byte> property to Options class, defaulting to ArrayPool<byte>.Shared.

### DIFF
--- a/UnitTests/Tests.cs
+++ b/UnitTests/Tests.cs
@@ -138,6 +138,7 @@ namespace Microsoft.IO.UnitTests
         }
 
         [Test]
+        [Ignore("ArrayPool")]
         public virtual void GetLargeBufferAlwaysAMultipleOrExponentialOfMegabyteAndAtLeastAsMuchAsRequestedForLargeBuffer()
         {
             const int step = 200000;
@@ -154,6 +155,7 @@ namespace Microsoft.IO.UnitTests
         }
 
         [Test]
+        [Ignore("ArrayPool")]
         public virtual void AllMultiplesOrExponentialUpToMaxCanBePooled()
         {
             const int BlockSize = 100;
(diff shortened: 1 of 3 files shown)


benmwatson:
Provocative indeed...not sure how I feel about this one.

Probably the best thing for me to do would be drop it into some test machines running Bing/XAP and see how it compares with the baseline. Those machines are currently being used to validate Bing on .NET 9, but once that's done, this would be an interesting experiment.

paulomorgado:
Being based on `ArrayPool<T>` or not, it would be great to have a shared `RecyclableMemoryStreamManager` that would be used on a parameterless `RecyclableMemoryStream`.

paulomorgado:
@benmwatson, were you able to test this?

benmwatson:
Not yet. I've been completely slammed for the last few months and RMS has been on the backburner except for some trivial things. Still want to try this, however...

paulomorgado:
How I understand you. 😄

paulomorgado:
Hi @benmwatson, is this worse pursuing?

PR #370, opened 2025-01-17
Add clarity for the large pool of buffers and the GetBuffer method pertaining to the .NET max array length

campbell-reid-displayr:
I found this sentence unclear, but after looking through the code, it made more sense. I thought it may be helpful to add context to clear up any misunderstandings.


campbell-reid-displayr:
@benmwatson

benmwatson (approved):

PR #406, opened 2025-11-28
Switch MemoryStreamDoubleDispose event level from Critical to Verbose

### LINKED ISSUE #394: Consider changing log severity of double dispose so not critical
I think to make the library closer to a drop in replacement to MemoryStream it would be good if double dispose wasn't logged as a critical, maybe information instead.

I understand the rationale for logging it and agree that it's useful, but given that double dispose is explicitly called out as being fine in the Stream docs, Asp.Net Core double disposes returned streams, and there's no negative impact in practice, I think critical seems like an overly severe categorisation.

See also https://github.com/dotnet/aspnetcore/issues/59514

I can manually filter/modify the logs for my app, but seems like something to consider changing in the library?

inklesspen1rus:
Multiple disposes are common thing in .NET. MSDN says that too:

> [Dispose()](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.component.dispose?view=net-8.0#system-componentmodel-component-dispose) can be called multiple times by other objects. ([link](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.dispose?view=net-8.0#system-io-stream-dispose))

> To help ensure that resources are always cleaned up appropriately, a [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) method should be idempotent, such that it's callable multiple times without throwing an exception. Furthermore, subsequent invocations of [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) should do nothing. ([link](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose))

> You should be safe to call it more than once, though you should probably avoid it if you can. (community, [link](https://stackoverflow.com/a/5306896/31094731))

Critical events may spuriously fill up event log with critical events while operations like that aren't even logical error.

So I switched it's event level to `Verbose`. I don't think there are reasons to use Information for events like that, when AspNetCore [disposes streams twice](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/issues/394).

Closes #394

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ You can optionally configure the `RecyclableStreamManager.ThrowExceptionOnToArra
 | -----|-------|-------------|
 | MemoryStreamCreated | Verbose | Logged every time a stream object is allocated. Fields: `guid`, `tag`, `requestedSize`, `actualSize`. |
 | MemoryStreamDisposed | Verbose | Logged every time a stream object is disposed. Fields: `guid`, `tag`, `allocationStack`, `disposeStack`. |
-| MemoryStreamDoubleDispose | Critical | Logged if a stream is disposed more than once. This indicates a logic error by the user of the stream. Dispose should happen exactly once per stream to avoid resource usage bugs. Fields: `guid`, `tag`, `allocationStack`, `disposeStack1`, `disposeStack2`. |
+| MemoryStreamDoubleDispose | Verbose | Logged if a stream is disposed more than once. This indicates a logic error by the user of the stream. Dispose should happen exactly once per stream to avoid resource usage bugs. Fields: `guid`, `tag`, `allocationStack`, `disposeStack1`, `disposeStack2`. |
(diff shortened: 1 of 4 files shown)


benmwatson:
I'm convinced. I just checked and internally, we don't even track this metric. Verbose makes sense. 

However, I think you should make one more modification. See code review comment.

benmwatson (changes requested):
The `Dispose(bool)` method call is grabbing a stack trace, and this will only make sense now if Verbose logging is enabled. Can you modify the code like this in `Dispose`?

```csharp
if (this.memoryManager.options.GenerateCallStacks && Events.Log.IsEnabled(EventLevel.Verbose))
...
```

benmwatson (approved):

jeroenhabets:
@benmwatson quick question: will this project still publish NuGet releases, any time soon? Including for this (minor) improvement?

PR #454, opened 2026-08-27
Update Nuget.config and DependaBot to use safe Microsoft NuGet proxy

benmwatson:
Safe feeds are now required. The default api.nuget.org feed is no longer accessible on Microsoft corporate devices.

NUnit3TestAdaptor needs to be switched back to 6.2.0 as that's the latest version available in this feed.


yihpink (approved):
lgtm

